PR #25, opened 2023-07-16
Fix for incompatible operator error

demoran23:
When attempting to install a new custom node, I received the following error:

> Install(git-clone) error: {git url} / '>=' not supported between instances of 'str' and 'int'

This fixed it for me


ltdrdata (approved):
Oh... my mistake.

PR #73, opened 2023-08-19
Update custom-node-list.json

dagthomas:
Added comfyui custom node for SDXL Auto Prompter


dagthomas:
Added custom SDXL Auto Prompter

PR #113, opened 2023-09-20
Change Clybius/ComfyUI-Latent-Modifiers's "install_type"

Xephobia:
The repo needs to be cloned
Changes `Clybius/ComfyUI-Latent-Modifiers`'s `install_type` from `copy` to `git-clone`


ltdrdata:
There was a minor mistake in your PR, so I made a separate correction and applied it. When changing to git-clone, "files" should also be changed to the repo URL.

PR #181, opened 2023-11-19
Update custom-node-list.json

noembryo:
Wrong files pulled..

### DIFF
--- a/custom-node-list.json
+++ b/custom-node-list.json
@@ -2452,7 +2452,7 @@
             "title": "noEmbryo nodes",
             "reference": "https://github.com/noembryo/ComfyUI-noEmbryo",
             "files": [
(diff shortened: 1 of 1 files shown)


ltdrdata:
Install type should be changed.
I just modified manually.

noembryo:
I think that the path is still wrong..
Shouldn't be `https://github.com/noembryo/ComfyUI-noEmbryo`?

ltdrdata:
> I think that the path is still wrong.. Shouldn't be `https://github.com/noembryo/ComfyUI-noEmbryo`?

That's right. I am modifying..

noembryo:
Thank you and sorry for the mess..  😳 
It won't happen again (hopefully 😃).

PR #207, opened 2023-12-02
add  EasyCaptureNode to custom-node-list

zhuanqianfish:
EasyCaptureNode allows you to capture any window, for later use in the ControlNet or in any other node.


ltdrdata:
This is registered already.
So I just reflected modified description.

PR #227, opened 2023-12-08
feat: added simple resolution selector node

TerminalCalm:
Requesting to add this custom node to the list.


ltdrdata:
You should not change the formatting of the entire file. This makes it difficult to identify which specific parts have been modified.

TerminalCalm:
Oops, my mistake, first time doing this. I'll withdraw and attempt to do it proper, must have been Prettier kicking in or something.

PR #242, opened 2023-12-15
add hayo comfyui coustom nodes tools

1shadow1:
Add hayo comfyui coustom nodes tool，Used to resize image formats, create transparent masks, directly load PIL type images from local and other functions


ltdrdata:
I merged this manually.

PR #270, opened 2023-12-27
Click "Workflow Gallery" shows the options

qiaoshun8888:
Also Fixed the "Close" menu behavior: it should close the menu popover instead of the entire Manager dialog

https://github.com/ltdrdata/ComfyUI-Manager/assets/1053081/926e8fe5-6195-48ca-923e-537806189e69

### DIFF
--- a/js/comfyui-manager.js
+++ b/js/comfyui-manager.js
@@ -851,9 +851,40 @@ class ManagerMenuDialog extends ComfyDialog {
 				$el("button", {
           id: 'workflowgallery-button',
           type: "button",
-          textContent: "Workflow Gallery",
-          onclick: this.handleWorkflowGalleryButtonClick,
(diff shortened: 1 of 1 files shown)


ltdrdata:
This seems to require some consideration. The primary buttons are intended to provide immediate access to frequently visited links, while the side buttons are designed for accessing other cases or scenarios.

In the popup menu opened by the side button, the checked state is displayed, and it seems more natural to open the last visited path. It might be beneficial to have an indication on the button itself regarding the link it is currently bound to, but it's something to consider.

qiaoshun8888:
https://github.com/ltdrdata/ComfyUI-Manager/assets/1053081/6e589c9e-1197-4562-91cf-95988bdc3ba2

PR #310, opened 2024-01-08
Fix notice board padding

shinshin86:
Hello!

I noticed that the padding in the notice board area didn't seem to be set up as planned, so I've made some adjustments. 
(Two paddings were set and one was disabled.)

Would appreciate it if you could take a look and let me know your thoughts.

Thank you!

## Screeshots
Before
fix-notice-board-padding_before

After
fix-notice-board-padding_after


ltdrdata:
When installing `https://github.com/11cafe/comfyui-workspace-manager`, there is an issue with the layout getting distorted. This needs to be taken into consideration.

shinshin86:
@ltdrdata Thanks for the comment!
I installed `comfyui-workspace-manager` and checked it out.
It seems that using workspace manager overwrites the styles. This may be a bit of a difficult issue here...

I'll keep looking for a solution!

ltdrdata:
> @ltdrdata Thanks for the comment! I installed `comfyui-workspace-manager` and checked it out. It seems that using workspace manager overwrites the styles. This may be a bit of a difficult issue here...
> 
> I'll keep looking for a solution!

Yeah. That is the problem.
Mainly, sytle confusing of `box-sizing: content-box;`  <-> `box-sizing: border-box;` is the problem.

ltdrdata:
The layout and size have recently changed a bit. Please adjust after syncing.

shinshin86:
@ltdrdata Apologies for the delay, but I have now made the necessary corrections.

Regarding the issue at hand, the direct cause appears to be the Chakra UI used by comfyui-workspace-manager, which seems to be affecting the overall style of ComfyUI. However, adjusting the scope of styles applied on the comfyui-workspace-manager side seemed to be quite challenging.

Therefore, I decided to apply additional styles to only the affected areas on our side, ensuring that the layout remains intact. This approach should solve the problem with minimal changes.

I have tested the styling with comfyui-workspace-manager both enabled and disabled to ensure that there are no layout issues.

Please let me know if you encounter any problems on your end.

Thank you!

shinshin86:
@ltdrdata 

> The layout and size have recently changed a bit. Please adjust after syncing.

OK! I'm merged and pushed the latest fixes! (https://github.com/ltdrdata/ComfyUI-Manager/pull/310/commits/27dd2a95fe79794cb7c9a7cdbba449a6dc59c801)

PR #423, opened 2024-02-21
Update custom-nodes-list.json

klinter007:
Added node for contact input strings

### DIFF
--- a/custom-node-list.json
+++ b/custom-node-list.json
@@ -5617,11 +5617,11 @@
             "description": "This is a node to convert an image into a CMYK Halftone dot image."
         },
 		{
-            "author": "Klinter",
+             "author": "Klinter",
             "title": "Klinter_nodes",
-            "reference": "https://github.com/klinter007/giling-around",
+            "reference": "https://github.com/klinter007/klinter_nodes",
             "files": [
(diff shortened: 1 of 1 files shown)


ltdrdata:
Since subfolders cannot be cloned, the `__init__.py` must be located at the root of the repository to enable a clonable structure.
Additionally, for custom nodes in the form of git-clone, the repository path should be included in the files.

ltdrdata:
Is it correct to provide it as a single file instead of the git clone method?

In this case, the file would be directly downloaded to the ComfyUI/custom_nodes path, and users cannot check for updates, so they cannot receive support for this feature.

If you are really going to provide it like that, the install_type should be set to `copy` instead of `git-clone`.

klinter007:
I have no idea which is the better option, these are my baby steps in the world of open source and esp. git.

I would like to add more things to it later, more nodes and probably more files with it.

ltdrdata:
Just change this
```
            "files": [
                "https://github.com/klinter007/klinter_nodes/blob/main/concat_klinter.py"
            ],
```
to

```
            "files": [
                "https://github.com/klinter007/klinter_nodes"
            ],
```

klinter007:
ok,
hope i've done it right 8-)

PR #451, opened 2024-03-02
Add ComfyUI-ergouzi-DGNJD to custom-node-list

11dogzi:
This is a multifunctional node kit that includes prompt word management, local refinement nodes, watermark addition, and other node groups. I have checked the format again and confirmed that my node can be opened in DB: Local mode. Thank you again for the outstanding contribution of ltdrdata

### DIFF
--- a/custom-node-list.json
+++ b/custom-node-list.json
@@ -1,6017 +1,4124 @@
 {
     "custom_nodes": [
         {
-            "extensions": [
-                {
-                    "author": "11dogzi",
-                    "title": "ComfyUI-ergouzi-DGNJD",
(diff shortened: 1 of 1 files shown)


ltdrdata:
1. Only additions to your custom node should be made, and no other modifications are allowed.
2. I request that the custom node, currently in Chinese, be translated into English. By default, ComfyUI-Manager aims to provide custom nodes that users worldwide can use.
   I recommend utilizing https://github.com/AIGODLIKE/AIGODLIKE-ComfyUI-Translation for multi-language support.
3. Please remove .git from the repository url you add. In other words, it should be (https://github.com/11dogzi/Comfyui-ergouzi-DGNJD).

11dogzi:
Received, I will make the necessary modifications before submitting

PR #483, opened 2024-03-14
Added Eagleshadow Custom Nodes entry to custom-node-list.json

Shadetail:
Added Eagleshadow Custom Nodes entry to custom-node-list

### DIFF
--- a/custom-node-list.json
+++ b/custom-node-list.json
@@ -6436,6 +6436,16 @@
             ],
             "install_type": "unzip",
             "description": "This is a node to convert an image into a CMYK Halftone dot image."
+        },
+        {
+            "author": "Shadetail",
+            "title": "Eagleshadow Custom Nodes",
+            "reference": "https://github.com/Shadetail/ComfyUI_Eagleshadow",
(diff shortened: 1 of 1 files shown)


ltdrdata:
Why don't you register as Is there a reason for registering through `copy` method instead of using `git-clone` method? Registering through copying does not allow providing update support to users.

Shadetail:
Oh wow thanks! I'm really glad you noticed that.

This is actually my first PR ever. I saw most other recent entries use `copy` so I thought that's what I was supposed to use as well. I didn't know that would prevent providing update support. Maybe this info could be added [to your readme.](https://github.com/ltdrdata/ComfyUI-Manager#how-to-register-your-custom-node-into-comfyui-manager)

I updated the entry so that it uses `git-clone` instead, as the ability to install and update nodes through manager is the main reason for adding entry for my nodes to manager.

ltdrdata:
You need to change 'files' field, too.
Change it to repo url instead of file url.

Shadetail:
Oops! Fixed!

PR #508, opened 2024-03-20
Update extension-node-map.json

julien-blanchon:


ltdrdata:
This is not updated manually but rather through scripts.

PR #528, opened 2024-03-30
Update custom-node-list.json

MokkaBoss1:
adding the MokkaBoss1 repository following requested ammendments:

1) remove pycache
2) update folder structure

### DIFF
--- a/custom-node-list.json
+++ b/custom-node-list.json
@@ -6959,30 +6959,7 @@
             ],
             "install_type": "copy",
             "description": "Extension to show random cat GIFs while queueing prompt."
-        },
-        {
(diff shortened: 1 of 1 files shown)


aurel-g:
The "reference" field for https://github.com/MokkaBoss1/ComfyUI_Mokkaboss1 points to the wiki, not the repository root. This is unusual; shouldn't this be fixed?

MokkaBoss1:
Hi,

likely I at the time I thought more people would want the explanations
about how to use each node rather than the actual code as this can be
easily installed from the Manager.

If enough people complain about this I'll switch it over.

thanks for pointing it out.

Kind regards

MokkaBoss1

PR #603, opened 2024-04-23
Minimal example to catch and show stderr. 

DrJKL:
Only accounts for one of the subprocesses.

I got caught with the `comfyui_manager_path` vs `custom_nodes_path` bug and the errors were in stderr, not stdout.
I'd have to switch back to the bad commit to reproduce it, let me know if you think that's valuable :-)


ltdrdata:
Oh... You are tryting to redirecting stderr messages.
I also think that would seem more informative.

DrJKL:
Want me to flesh it out and do the rest of the processes?

ltdrdata:
If you PR it. I'll merge it.

PR #606, opened 2024-04-24
Fix the bug of "Unload models"

ty0x2333:
The `/free` API of comfyanonymous/ComfyUI has changed. Commit: https://github.com/comfyanonymous/ComfyUI/commit/6d281b4ff4ad3918a4f3b4ca4a8b547a2ba3bf80

Now, the `/free` API requires `unload_models` and `free_memory` parameters, and the default value is False.

https://github.com/comfyanonymous/ComfyUI/blob/master/server.py#L509

Currently the "Unload Models" button no longer works. It looks like this bug has been around for a long time.

This PR updates the `/free` API call and fixes this BUG.

### DIFF
--- a/js/common.js
+++ b/js/common.js
@@ -154,7 +154,7 @@ export async function free_models() {
 	let res = await api.fetchApi(`/free`, {
 						method: 'POST',
 						headers: { 'Content-Type': 'application/json' },
-						body: '{}'
+						body: '{"unload_models": true}'
 					});
 
 	if(res.status == 200) {


Thireus:
Thanks for pointing this out! Doesn't also require free_memory set to True now as well? Why not free the memory along with unloading the models? The button doesn't seem to do much with Flux as mentioned here: https://github.com/comfyanonymous/ComfyUI/issues/4217

ltdrdata:
> Thanks for pointing this out! Doesn't also require free_memory set to True now as well? Why not free the memory along with unloading the models? The button doesn't seem to do much with Flux as mentioned here: [comfyanonymous/ComfyUI#4217](https://github.com/comfyanonymous/ComfyUI/issues/4217)

image

There are two kinds unload button.
This is only available via new style ui.

Thireus:
Oh I see, I'm still using the legacy UI I believe because this is what I have:

unload_model

PR #676, opened 2024-05-13
fix deps of comfyui bedrock

yytdfc:


ltdrdata:
You don't need specifying pip in there since they are specified in `requirements.txt` already.
The purpose is hotfix.

PR #689, opened 2024-05-17
add a new node about ID_Animator

smthemex:
ID_Animator, using  zero shot method


ltdrdata:
This is added, already.
I just updated the description.

PR #700, opened 2024-05-21
Update custom-node-list.json

ppajkasd:
change hfdownload node id


ltdrdata:
this is reflected manually.

PR #744, opened 2024-06-05
New custom nodes manager

cenfun:
## Motivation
The purpose of this PR is to provide a better custom node manager.

## Optimizations and improvements
- Performance optimization. Currently, we have nearly a thousand custom nodes, and there are performance issues, especially when selecting and scrolling can cause the browser to freeze. This update will use a high-performance grid to render the list.

- Eliminate the duplicate page `a1111-alter-downloader`, and merge it to new custom node manager (additional column).

- High-performance real-time keyword search, better user experience. 

- New filter and multi-select features, and so on.

cn

## TODO
I hope someone can help test this PR locally, please let me know if there are any issues and your questions. Thanks.


ltdrdata:
After clicking the Custom Nodes Manager button, it hangs.

image
image
image

cenfun:
@ltdrdata  thanks for the feedback, some of browser version may not support build-in crypto, I just replace it with md5, please try again.

ltdrdata:
It seems the 'select all' checkbox is not functioning properly. 

The original behavior is that when you check one item and then 'select all', it should select all nodes of the same kind that can have the same action applied within the current filter view. 

And when in the 'select all' state, it should deselect them all.

ltdrdata:
image

It would be better if the link in the description was more clearly distinguishable.

cenfun:
> It seems the 'select all' checkbox is not functioning properly. The original behavior is that when you check one item and then 'select all', it should select all nodes of the same kind that can have the same action applied within the current filter view. And when in the 'select all' state, it should deselect them all.

Yes the behavior of `select-all` is different from original. I think the original calculation logic is very complicated. The new logic is to group all selections, rather than being able to only select the same kind of type. Does this make sense?

> It would be better if the link in the description was more clearly distinguishable.

how about increase the font size and line height a bit or any color?

ltdrdata:
> > It seems the 'select all' checkbox is not functioning properly. The original behavior is that when you check one item and then 'select all', it should select all nodes of the same kind that can have the same action applied within the current filter view. And when in the 'select all' state, it should deselect them all.
> 
> Yes the behavior of `select-all` is different from original. I think the original calculation logic is very complicated. The new logic is to group all selections, rather than being able to only select the same kind of type. Does this make sense?
> 
> > It would be better if the link in the description was more clearly distinguishable.
> 
> how about increase the font size and line height a bit or any color?

I see you've organized it so that 'select all' is the default when nothing is selected.

ltdrdata:
The nodes should be listed as requiring updates when commit of some custom nodes are deleted using `git reset HEAD~1 && git reset --hard`, and then `skip update check` option is disabled.


Issues:
1. Updatable items not reflected immediately. (maybe cache issue?)

2. `skip update check` Misapplication: It seems the skip update check option is not being applied correctly. 
(This option is crucial in environments with unstable cloud or network conditions.)

3. Disappearing Update Button: After restarting, the update button appears correctly when opening the first dialog. However, upon closing and reopening the dialog, the button disappears.

cenfun:
> The nodes should be listed as requiring updates when commit of some custom nodes are deleted using `git reset HEAD~1 && git reset --hard`, and then `skip update check` option is disabled.
> 
> Issues:
> 
> 1. Updatable items not reflected immediately. (maybe cache issue?)
> 2. `skip update check` Misapplication: It seems the skip update check option is not being applied correctly.
>    (This option is crucial in environments with unstable cloud or network conditions.)
> 3. Disappearing Update Button: After restarting, the update button appears correctly when opening the first dialog. However, upon closing and reopening the dialog, the button disappears.

In the new implementation, I have not changed any APIs, just added a API `/customnode/alternatives`.
I'm not very familiar with the existing APIs, but following are my understanding of the APIs and implementation of loading data list:

- ShowMode.Normal: `/customnode/getlist?mode=${mode}&skip_update=true` 
- ShowMode.Update: `/customnode/getlist?mode=${mode}` 
- ShowMode.Missing: merge `/customnode/getlist?mode=${mode}&skip_update=true`  with `/customnode/getmappings?mode=${mode}` + `/component/get_unresolved`
- ShowMode.Alternatives: merge `/customnode/getlist?mode=${mode}&skip_update=true`  with `/customnode/alternatives?mode=${mode}`

see function [loadData](https://github.com/cenfun/ComfyUI-Manager/blob/custom-nodes-manager/js/custom-nodes-manager.js#L1205)

(Check Update = ShowMode.Update, Check Missing = ShowMode.Missing)
Should we add a button `Fetch Updates` in new manager dialog?

And following are `install` buttons APIs when clicking:

- `Enable` and `Disable`:  `/customnode/toggle_active`
- `Update` and `Try update`: `/customnode/update`
- `Try fix`: `/customnode/fix`
- `Install` and `Try install`:  `/customnode/install`
- `Uninstall`: `/customnode/uninstall`

see [getInstallButtons](https://github.com/cenfun/ComfyUI-Manager/blob/custom-nodes-manager/js/custom-nodes-manager.js#L396) and [installNodes](https://github.com/cenfun/ComfyUI-Manager/blob/custom-nodes-manager/js/custom-nodes-manager.js#L902)

Please correct if there are anything wrong. or should I edit the existing APIs or add new APIs? Thanks.

> Disappearing Update Button: After restarting, the update button appears correctly when opening the first dialog. However, upon closing and reopening the dialog, the button disappears.

BTW, the issue 3 should be fixed.

ltdrdata:
Upon further investigation, I found that when the dialog is opened for the first time, the API call is correctly made with `skip_update=true`. However, at the moment when the filter is changed to update, the API is called again without `skip_update=true` being applied.

Actually, under the existing policy, I would have said that there's no need to make additional API calls before opening a new dialog, but I am a bit concerned because I plan to introduce the concept of paging in the future.

In any case, setting `skip_update` to false is still a burdensome situation, so it needs to be applied appropriately.

The worst-case scenario when `skip_update` is disabled is that if user use a colab with ComfyUI installed in gdrive, it takes more than 10 minutes due to accessing the git repos of entire custom nodes. This situation occurs as the git repo on the gdrive, which normally wouldn't be accessed, is being cached through the network to the Colab instance.

cenfun:
@ltdrdata  Thanks for the feedback.
I also thought that the updates should not be checked at the first time, because it may take long time. 

For pagination, I personally think it's a bad idea because it offers a poor user experience. such as, during a search or sort, if my implementation is used, there's absolutely no need for pagination. The grid I'm using can support rendering tens of thousands of rows of data with high-performance.

- try to speed up installation by using a shallow clone: `git clone --depth=1 <url>`  see [here](https://github.blog/2020-12-21-get-up-to-speed-with-partial-clone-and-shallow-clone/)
- Could we check the updates one by one instead of checking all at once, causing the process to block? (need a single update checking API), Or, support not checking again within 10 minutes after a successful check is completed.

Are there any other issues about this PR?

ltdrdata:
I am currently upgrading to the new node installation system. 
Instead of relying primarily on Git, it will be structured to primarily support CNR(Comfy Registry) instead of Git. 

This doesn't mean I am removing the current Git-based installation; it will still be available but considered as a nightly installation without semantic versions. 

Once the CNR-based node installation system is established, I expect many performance-related issues to be significantly mitigated.

ltdrdata:
image

I added marker for author.
https://github.com/ltdrdata/ComfyUI-Manager/blob/44ad47709719a605c4306d089daf074ea52a51f4/js/custom-nodes-downloader.js#L570

cenfun:
merged and updated
image

ltdrdata:
This error occurs when the filter is changed to a missing node.
(For reference, this occurs when the missing node does not exist.)
image

cenfun:
@ltdrdata It should be fixed, thanks for the feedback.

ltdrdata:
This is merged. Thx.

PR #771, opened 2024-06-14
Update custom-node-list.json

baicai99:
This node provides frame control, such as: save the first 16 images, ignore the first 16 images, ignore the first 16 masks, and so on.


baicai99:
I'm a newbie. I don't know if this is the right submission.

ltdrdata:
Remove `__pycache__` dir from the repo.
That should not exists in the repo.

PR #778, opened 2024-06-15
Comfyui_Bxb

zhulu111:


ltdrdata:
Due to security concerns, obfuscated code cannot be registered with ComfyUI-Manager.

PR #780, opened 2024-06-16
update DB

zhulu111:


ltdrdata:
Don't use Chinese characters in the description and node names/widgets.
For translations, please apply them through https://github.com/AIGODLIKE/AIGODLIKE-ComfyUI-Translation.

PR #802, opened 2024-06-21
Better model manager UI

cenfun:
This PR is to update a better model manager UI.
(There is no API changes, only UI changes)

## Optimizations and improvements
- New model manager UI (see below screenshot)
- `Install Models` menu is changed to `Model Manager`
- Model filter for `type` and `base`
- Minor optimizations for `Custom Nodes Manager`
- Updated grid lib `turbogrid.esm.js` to latest version

cmm


ltdrdata:
How about applying this animation to the model download as well?

image

cenfun:
> How about applying this animation to the model download as well?

done, please review again.

PR #809, opened 2024-06-22
Supports model size

cenfun:
- Update `model-list.json` for size data.
- Show size column for model list.

image

How to get the model size?
Sending `HEAD` request to the model `url`, and get `content-length` from the reponse header.


ltdrdata:
oh.. wow.. You did a lot of work.

PR #810, opened 2024-06-22
add ComfyUI_Zwng_Nodes

za-wa-n-go:
Simple nodes for loading image files.Nodes that include a simple remote connection to Photoshop, a node that can overlay and preview an image with a mask, and a node that can load images directly from a file path.


ltdrdata:
Change
"https://github.com/za-wa-n-go/ComfyUI_Zwng_Nodes/blob/main/zwng_nodes.py"

to

"https://github.com/za-wa-n-go/ComfyUI_Zwng_Nodes"

ltdrdata:
This is added manually due to incorrect PR.

PR #831, opened 2024-07-02
Adding Copus to the sharing options list

copus-dev09:
### Sharing options list
image

### Sharing window
image


ltdrdata:
1. I entered the API key and shared an image once, but after rebooting ComfyUI and refreshing, I noticed that the API key was cleared. A patch is needed to make the API key stored.

2. On the link accessed through "Get API Key," the API key cannot be seen immediately. After visiting the site, I have to search around to find the page where I can get the API Key.

PR #851, opened 2024-07-10
Add LM Studio Image to Text Node to custom node list

mattjohnpowell:
Add LM Studio Image to Text Node to custom node list
A simple node to connect to LM Studio running on localhost for image classification


ltdrdata:
Remove the `__pycache__` directory from the repo, and register it using `git-clone` instead of `copy`.
The copy method should generally not be used due to issues such as the inability to track updates.

ltdrdata:
This is added manually.

PR #860, opened 2024-07-12
Add ComfyUI-FrameFX to custom node list

mgfxer:
A collection of custom nodes for ComfyUI that enhance video and animation workflows. Features include:
Slideshow creation, Animated transitions, Automated mask code generation, Text-based prompt travel, AI-assisted prompt travel generation using Blip or Florence captions also works with any double spaced text, Live prompting integration for custom zoom workflows using auto-queue.
Includes custom workflows that use the nodes.


ltdrdata:
Your node is added manually, already.
I just updated description.

PR #879, opened 2024-07-17
node name fixing for ComfyUI-Inspyrenet-Rembg

john-mnz:
the node could not be found by "install missing custom nodes" so I fixed the name here


ltdrdata:
That file is managed by script.

PR #886, opened 2024-07-19
added support for `--no-deps` option to node install and reinstall

telamonian:
Adds support for a `no-deps` option to the `node install` and `node reinstall` commands. Not only is this useful to have in general, but it's needed to support the new `--fast-deps` behavior (unified/much faster python dependency installs) in comfy-cli.

Two unresolved questions:
- What commands aside from install/reinstall should also get a `no-deps` option? I added plumbing for  `no-deps` such that it should be respected for any command that sets `cm_ctx.no_deps = True`, so it should now be easy to add to any command
- I noticed that adjacent to `requirements.txt` support, there is also support for running some kind of generic `<node_root>/install.py` installation script. Should this installation script support also be disabled by the `no-deps` option? For now I've left it untouched

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,8 @@
 __pycache__/
 .idea/
 .vscode/
+.history/
+*.code-workspace
 .tmp
 .cache
 config.ini
--- a/cm-cli.py
+++ b/cm-cli.py
@@ -80,6 +80,7 @@ read_downgrade_blacklist()  # This is a preparation step for manager_core
 class Ctx:
     def __init__(self):
         self.channel = 'default'
+        self.no_deps = False
         self.mode = 'cache'
 
     def set_channel_mode(self, channel, mode):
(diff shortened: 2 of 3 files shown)


robinjhuang:
@telamonian Can you make this to the CNR feature branch? It has diverged too far from main so for that to merge easily, it would be prefer to have this PR go to that branch.
branch name: `feat/cnr`

telamonian:
@robinjhuang The PR has now been rebased on top of `feat/cnr`

@ltdrdata The dependency handling and other code in this PR appears to be working, but I am getting an unrelated error when I try to install any custom node:

```bash
comfy --here node install kjnodes

...

ERROR: An error occurred while installing 'kjnodes'.
Node 'kjnodes@unknown' not found in 
```

I get this same error when using just the base `feat/cnr` branch, which is why I don't think it has anything to do with this PR's code.

ltdrdata:
> @robinjhuang The PR has now been rebased on top of `feat/cnr`
> 
> @ltdrdata The dependency handling and other code in this PR appears to be working, but I am getting an unrelated error when I try to install any custom node:
> 
> ```shell
> comfy --here node install kjnodes
> 
> ...
> 
> ERROR: An error occurred while installing 'kjnodes'.
> Node 'kjnodes@unknown' not found in 
> ```
> 
> I get this same error when using just the base `feat/cnr` branch, which is why I don't think it has anything to do with this PR's code.

The id is `comfyui-kjnodes` not `kjnodes`.
Unfortunately there is no short alias in CNR, yet.

You can use `comfy node show cnr` in comfy-cli's `dr-feat-cnr` branch.

PR #895, opened 2024-07-22
Add FAI-Node custom nodes

alanhuang67:
Add FAI-Node custom nodes

### DIFF
--- a/custom-node-list.json
+++ b/custom-node-list.json
@@ -12164,6 +12164,17 @@
             "install_type": "unzip",
             "description": "Various image processing nodes."
         },
+		{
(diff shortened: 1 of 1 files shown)


ltdrdata:
1. Remove `nodes/__pycache__` and `modules/__pycache__` from repo.
(I recommend add them into .gitignore as well)
2. Do not use Chinese for widget values. By default, we aim for global users. For multi-language support, it is recommended to provide it through https://github.com/AIGODLIKE/AIGODLIKE-ComfyUI-Translation.

alanhuang67:
Done, plz check, thanks~

PR #911, opened 2024-07-29
Remove duplicate SAM models from the model-list

punitda:
The [model-list.json](https://github.com/punitda/ComfyUI-Manager/blob/main/model-list.json) file contains duplicate SAM models in the list. We can remove the duplicates where we use this models list in our code. However, IMO it makes sense to clean up the source itself


ltdrdata:
oh... thank you.

PR #922, opened 2024-08-02
Update custom-node-list.json

PowerHouseMan:


ltdrdata:
Basically, ComfyUI-Manager aims to support international users.

Korean was used in the display name of the node, but it is recommended to remove this and use the following translation extension for Korean support: https://github.com/AIGODLIKE/AIGODLIKE-ComfyUI-Translation

PowerHouseMan:
I removed the Korean characters. Please review again.

PR #936, opened 2024-08-06
ComfyUI-SP-Nodes

bananasss00:
Node Pack: PromptChecker for token toggling, KoboldCPP API, ModelMerging, Telegram-Bot-API, and more


ltdrdata:
Nodes that carry the risk of executing arbitrary code based on workflow data cannot be registered in the default node list due to security policy.

In such cases, they are provided in the dev list with an [UNSAFE] mark, along with an explanation of the associated risks.

bananasss00:
Removed unsafe node from pack

PR #942, opened 2024-08-09
 feat: mapping clip and checkpoint to model path

allmelgr:
Hi Ltdrdata,

I seen some change in model-list.json:
   the type "checkpoints" was changed to "checkpoint", but in the glob.manager_server.get_model_dir without this mapping
   and comfyanonymous/clip_l is default path, clip is also without mapping

And I seen the defualt path is **etc**, not in ComfyUI/models/etc. maybe use `os.path.join(folder_paths.models_dir, "etc")` is better?

### DIFF
--- a/glob/manager_server.py
+++ b/glob/manager_server.py
@@ -249,8 +249,12 @@ def get_model_dir(data):
         model_type = data['type']
         if model_type == "checkpoints":
             base_model = folder_paths.folder_names_and_paths["checkpoints"][0][0]
+        elif model_type == "checkpoint":
+            base_model = folder_paths.folder_names_and_paths["checkpoints"][0][0]
         elif model_type == "unclip":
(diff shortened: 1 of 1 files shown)


ltdrdata:
oh.. I have missed that.

jamesbrink:
Thank you for this fix. I was pulling my hair out wondering why some models were going into `./etc`

allmelgr:
> Thank you for this fix. I was pulling my hair out wondering why some models were going into `./etc`

I mapped the checkpoint and clip. This doesn't fix the issue with `./etc`, but I suggest using `ComfyUI/models/etc` instead of `./etc`

ltdrdata:
> > Thank you for this fix. I was pulling my hair out wondering why some models were going into `./etc`
> 
> I mapped the checkpoint and clip. This doesn't fix the issue with `./etc`, but I suggest using `ComfyUI/models/etc` instead of `./etc`

oops...

PR #966, opened 2024-08-15
Add RTX Remix custom nodes

nxkb:
Hello!
This PR adds our new nodes for integration with the RTX Remix App.

I tested the install using the local DB, and it worked! One question though is whether it should have recognized the missing nodes already or if that will only happen after this merges?

Let me know if there's something else to do on my side. Cheers.
Nicolas


ltdrdata:
Node mapping information will be collectively updated by a script on my end after the merge.

PR #995, opened 2024-08-23
Adding ComfyUI-LoadFiles

daryltucker:
1. Added [ComfyUI-LoadFiles](https://github.com/daryltucker/ComfyUI-LoadFiles)
2. Removed extra newlines from `custom-node-list.json`
3. Loaded Manager w. "DB: Local" and tested display.

> The primary goal of these nodes is to provide a way to access files generated by ComfyUI workflows, local to the machine running ComfyUI.  These nodes should always return an updated list of files when triggered.

Thank you


ltdrdata:
This is added manually because of conflict.

PR #1027, opened 2024-09-01
Faster plugins installation & less space taken

### LINKED ISSUE #891: [Feature Request] Clone depth
Please add possibility to clone only 1 last version (--depth 1) by option (or by default)
There are too much unneeded files/space used and time when cloning objects.
Some of the nodes takes x50-x100 space and a lot of MFT records for nothing never be needed.

KaruroChori:
Closes https://github.com/ltdrdata/ComfyUI-Manager/issues/891
This change allows ComfyUI-Manager to only clone the latest commit from the relevant branch, so that we do not end up with some extensions taking up GB on disk.
Yes, this is a real issue :(.

### DIFF
--- a/glob/manager_core.py
+++ b/glob/manager_core.py
@@ -563,11 +563,11 @@ def gitclone_install(files, instant_execution=False, msg_prefix=''):
 
             # Clone the repository from the remote URL
             if not instant_execution and platform.system() == 'Windows':
-                res = manager_funcs.run_script([sys.executable, git_script_path, "--clone", custom_nodes_path, url], cwd=custom_nodes_path)
+                res = manager_funcs.run_script([sys.executable, git_script_path, "--clone --depth=1 --single-branch --recurse-submodules", custom_nodes_path, url], cwd=custom_nodes_path)
                 if res != 0:
                     return False
             else:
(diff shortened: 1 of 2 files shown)


ltdrdata:
Due to the current overhaul, this PR will be considered after that.

drmbt:
are we at a place where this could be considered yet?

ltdrdata:
Oh, this will be reflected after switching to the pip version later.

KaruroChori:
I mean, let's be real. Those four lines changed were quite literally just allowing the extension to limit cloning to the master/main branch, and ignoring most of the git history, only taking the last commit.  

At times people do mess up with their repos, like they introduce very big files which should have never been there, and remove them right after without amending a commit. So that big file is part of the history and one is forced to take it unless they limit the depth of cloning.

It is really not rocket science, I don't understand why any rewrite would/should have prevented this feature from shipping for half a year, but if anyone is interested it can be patched in quite easily on your systems.

ltdrdata:
There are more side effects to consider than expected. 
In particular, the snapshot feature is designed to fetch a specific commit, but if only the latest commit is fetched, it can cause issues—so handling edge cases becomes necessary.

Also, applying a patch that touches various parts of the code in the current version could mess up the branch currently being worked on, where the code structure is being completely overhauled.

PR #1032, opened 2024-09-03
add custom node info

MoonHugo:
add custom node info


ltdrdata:
When adding a node, you should not touch any other parts. Only the part you added should be changed.

PR #1034, opened 2024-09-05
add custom node info

MoonHugo:
add custom node info


ltdrdata:
This is added manually.

PR #1097, opened 2024-09-29
Update install_type for ComfyUI-Molmo in custom-node-list.json

CY-CHENYUE:
Changed the install_type for ComfyUI-Molmo from 'git_clone' to 'git-clone' to resolve installation issues and maintain consistency with other extensions.

The nodes are already available in the manager, but the installation prompt says “ComfyUI-Molmo install failed: Bad Request”, is this related?


CY-CHENYUE:
Issues arising
image

PR #1106, opened 2024-10-02
add nilor-nodes to custom node list db

stephennilor:
hey please accept adding of our nodes to the db 👍


stephennilor (commented):
👍 noted, thanks

PR #1122, opened 2024-10-07
Update custom-node-list.json

SS-snap:


ltdrdata:
Remove the `__pycache__`  dirs from repo.
Don't forget the subdirectories.

And if possible, it would be good to remove the `.idea` folder as well.

SS-snap:
Thanks~

PR #1156, opened 2024-10-17
add ComfyUI-Fluxtapoz to custom-node-list

logtd:
Add Fluxtapoz to custom node list


logtd:
Woops, my linter went crazy. Let me fix that

PR #1181, opened 2024-10-23
Update custom-node-list.json

ComfyPlus:
Add extension: ComfyPlus_Anywhere


ltdrdata:
This node registration is pending due to native code included in the `libs`.  
Currently, our policy does not allow code that contains unverifiable components.  
It may be considered in the future if a verification system for the author is established.

PR #1229, opened 2024-11-10
Update custom-node-list.json

Isi-dev:
Please add these animation nodes and workflows.


ltdrdata:
I recommend `.vscode` dir from the repo.
That easily corrupts the git state.

Isi-dev:
> I recommend `.vscode` dir from the repo. That easily corrupts the git state.

.vscode` dir has been removed.

PR #1253, opened 2024-11-20
manager_core: Fix fetch/update line erasure output issues

clayne:
* __win_check_git_update: Properly cleanup line before using a carriage return to return to the beginning. This prevents output like:
```
  Updated: S:\ComfyUI\custom_nodes\ComfyUI-Impact-Packrpolation
  Updated: S:\ComfyUI\custom_nodes\comfyui-inpaint-nodesayout
  Updated: S:\ComfyUI\custom_nodes\ComfyUI-LogicUtilsodes
```
  In this case the node being updated overlaps the existing output showing nodes being checked for updates. This happens because there's no erase until end of line being used in the windows specific function whereas the non-windows version uses it in git_repo_has_updates.

### DIFF
--- a/glob/manager_core.py
+++ b/glob/manager_core.py
@@ -367,13 +367,13 @@ def __win_check_git_update(path, do_fetch=False, do_update=False):
     if do_update:
         if "CUSTOM NODE PULL: Success" in output:
             process.wait()
-            print(f"\rUpdated: {path}")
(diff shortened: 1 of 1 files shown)


clayne:
Same approach used here: https://github.com/ltdrdata/ComfyUI-Manager/pull/1253/files#diff-072d070db24c8b304f2499fd9fd83954a35723cbf60fbcf82436837cca42255eL489

PR #1254, opened 2024-11-21
Add  ComfyUI-ascii-art custom node to the list

tomudo:
https://github.com/tomudo/ComfyUI-ascii-art

screenshot


ltdrdata:
There's no major issue, so I'll proceed with the merge, but it might be better to delete or modify the metadata section that was copy-pasted on this side.

https://github.com/tomudo/ComfyUI-ascii-art/blob/ea543989edd1f15046e6fa3a1fb7c8a34a7f6388/__init__.py#L4

PR #1264, opened 2024-11-24
Spelling Error...

kaibioinfo:
I'm really sorry, but somebody told me that its "Redux", not "Reflux". I think it's a bit too annoying to change the name of the plugin now - I like Reflux more anyways. But I would like to update the description such that people can find the plugin via search.

Sorry for the inconvenience.


ltdrdata:
This update is manually applied.

PR #1272, opened 2024-11-26
Add exectails custom nodes

exectails:
Adds multiple node repositories to the custom nodes list.


ltdrdata:
This node cannot be allowed in default channel.
https://github.com/exectails/comfyui-et_scripting

exectails:
Got it, makes sense^^ Thanks for merging the others.

PR #1278, opened 2024-11-28
feat: add  ComfyUI MiaoShua Creator

motivated3:


motivated3:
add comfyui-shua-creator

PR #1296, opened 2024-12-04
add Fix BiRefNet-utils

hieuck:
I'm newbie and don't know how to update custom-node. If there's anything that needs to be fixed, please let me know


ltdrdata:
For now, a temporary deferment is needed to see if the original author will continue maintaining it.

Since it appears to have been unmaintained for a long time, if your PR is not merged within about a week, I will replace it with your node.

PR #1300, opened 2024-12-05
Fix the situation that requiresments.txt contains `~=`

sko00o:
Reference to https://packaging.python.org/en/latest/specifications/version-specifiers/#id5


sko00o:
Related to #1301

sko00o:
rebased in https://github.com/ltdrdata/ComfyUI-Manager/pull/1439

PR #1303, opened 2024-12-06
Bypass ComfyUI version check when in electron env

webfiltered:
#### User changes
- Fixes bug - version too outdated & not in repo warnings are never displayed
- When using desktop, ComfyUI version check is skipped (handled by app)

#### Code / dev changes
- Detects Electron by via environment var (set by Electron itself)
- If debugging in an IDE built with Electron, this value will very likely be set


webfiltered:
@ltdrdata Can you please check [Fix exception when comfyui is not repo](https://github.com/ltdrdata/ComfyUI-Manager/commit/aedec50418a9f7045e4742274c3c08402453a403)?  I am not 100% that this won't break other things!  I can do it another way if it's an issue.  😄

webfiltered:
GitHub Desktop strikes again.  Wrong repo.  Feel free to close this if not relevant here.

PR #1353, opened 2024-12-19
Add extension to attach node versions

huchenlei:
This PR adds an extension that attaches node versions metadata to the serialized workflow. This will help improve the reproducibility of the workflow.

Only nodes in the graph will be included in the metadata.

### DIFF
--- a/js/workflow-metadata.js
+++ b/js/workflow-metadata.js
@@ -1,0 +1,72 @@
+/**
+ * Attaches metadata to the workflow on save
+ * - custom node pack version to all custom nodes used in the workflow
+ */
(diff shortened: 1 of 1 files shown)


js/workflow-metadata.js
```diff
@@ -0,0 +1,84 @@
+/**
+ * Attaches metadata to the workflow on save
+ * - custom node pack version to all custom nodes used in the workflow
+ *
+ * Example metadata:
+ "extra": {
```
yoland68, line 6: Is there a place where we have a list of definition for workflow fields?

PR #1354, opened 2024-12-20
Add Image Metadata Nodes by Light-x02

Light-x02:
This pull request adds a new custom node package called Image Metadata Nodes. It includes two complementary nodes for loading and saving images while preserving metadata intact.

**Features**
Image Metadata Loader: Loads an image and extracts its metadata.
Image Metadata Saver: Saves an image with its metadata unchanged, directly embedded in the PNG file.

**Use Case**
These nodes are particularly useful for workflows involving modifications like upscaling, where preserving metadata is essential. Users can load an image, pass its metadata through the workflow, and save the final output with intact metadata.

**GitHub Repository**
https://github.com/Light-x02/ComfyUI-Image-Metadata-Nodes

Thank you for considering this addition to ComfyUI Manager!


ltdrdata:
This is adde manually.

PR #1358, opened 2024-12-20
[CNR] Remove `@<version>` suffix on module name

huchenlei:
The biggest issue on the `<node>@<version>` folder (module) name is that it breaks all imports of that module using the un-suffixed name. There are plenty of such usage in the wild that cannot be easily fixed.

image

This PR removes the `@<version>` suffix from module name.

WIP (Needs verification from @ltdrdata)
The cache/switching version system needs a redesign.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,4 @@ github-stats-cache.json
 pip_overrides.json
 *.json
 check2.sh
+/venv/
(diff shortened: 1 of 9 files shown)


ltdrdata:
This PR is base of `feat/cnr-no-suffix`.

PR #1393, opened 2025-01-03
Fix NameError in get_custom_nodes_paths method

yhayano-ponotech:
## Summary  
Fixed an error in the `get_custom_nodes_paths` method where the `folder_paths` module could not be found.

## Changes Made  
- Added a class variable `folder_paths` to the `Ctx` class.  
- Dynamically imported the `folder_paths` module in the `__init__` method.  
- Updated the `get_custom_nodes_paths` method to use the class variable `folder_paths`.

## Reason for the Fix  
The `get_custom_nodes_paths` method raised a `NameError: name 'folder_paths' is not defined` because the `folder_paths` module was not found. This fix properly imports the module and resolves the error.

## Test Results  
After the fix, the `cm-cli.py` script worked as expected. Specifically, the `restore-dependencies` command no longer encountered errors.  

This improvement enhances the stability and reliability of the ComfyUI Manager.  
Your review would be greatly appreciated.

### DIFF
--- a/cm-cli.py
+++ b/cm-cli.py
@@ -12,6 +12,7 @@ from rich import print
 from typing_extensions import List, Annotated
 import re
 import git
+import importlib
 
 
 sys.path.append(os.path.dirname(__file__))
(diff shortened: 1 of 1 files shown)


unexpectedly:
Experiencing this issue in colab. Fix appreciated!

PR #1420, opened 2025-01-09
Update custom-node-list.json

ciga2011:
Add ComfyUI-AppGen


ltdrdata:
I think iframe of the sandbox node is not sufficiently isolating scripting, so it doesn't seem safe.

ciga2011:
got it.

PR #1424, opened 2025-01-09
Update custom-node-list.json

GraftingRayman:
a new node repository for PuLID with extended features not available on other versions


ltdrdata:
This is added already.

PR #1454, opened 2025-01-15
Update install-comfyui-venv-linux.sh

dsigmabcn:
git clone is now done in 'comfyui-manager'. pip install of the requirements of the manager needs to change the path name accordingly

### DIFF
--- a/scripts/install-comfyui-venv-linux.sh
+++ b/scripts/install-comfyui-venv-linux.sh
@@ -6,7 +6,7 @@ python -m venv venv
 source venv/bin/activate
 python -m pip install torch torchvision torchaudio --extra-index-url https://download.pytorch.org/whl/cu121
 python -m pip install -r requirements.txt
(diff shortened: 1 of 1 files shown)


scripts/install-comfyui-venv-linux.sh
```diff
@@ -6,7 +6,7 @@ python -m venv venv
 source venv/bin/activate
 python -m pip install torch torchvision torchaudio --extra-index-url https://download.pytorch.org/whl/cu121
 python -m pip install -r requirements.txt
-python -m pip install -r custom_nodes/ComfyUI-Manager/requirements.txt
+python -m pip install -r custom_nodes/comfyUI-manager/requirements.txt
```
ltdrdata: That should be fully lower case.

dsigmabcn: correct, made the change

ltdrdata (approved):

PR #1465, opened 2025-01-18
Add ImageSimilarity to custom-node-list.json

ngosset:
Just published my first custom node, looking to add it to custom-node-list.json 

Thanks!

PR #1466, opened 2025-01-18
Update custom-node-list.json

ngosset:


ngosset:
2nd PR was created by mistake, #1465  is correct PR

PR #1500, opened 2025-01-30
Fix import collision

ProGamerGov:
Currently the extension directory and dependency have the same name, and that leads to errors when trying to load it.

This PR fixes the issue.


ltdrdata:
You need to remove this.
https://registry.comfy.org/nodes/pytorch360convert

ProGamerGov:
@ltdrdata I have requested its removal here: https://github.com/Comfy-Org/registry-backend/issues/157 as I can't figure how to get the bearer token required to delete nodes.

PR #1501, opened 2025-01-31
fix channel_url config

956237586:
1.fix channel_url not effecte for default_cache_update
2.support http channel url for airgap env

### DIFF
--- a/glob/manager_core.py
+++ b/glob/manager_core.py
@@ -1625,7 +1625,7 @@ def get_config():
     if cached_config is None:
         cached_config = read_config()
         if cached_config['http_channel_enabled']:
-            print(f"[ComfyUI-Manager] Warning: http channel enabled, make sure server in secure env")
+            print("[ComfyUI-Manager] Warning: http channel enabled, make sure server in secure env")
 
     return cached_config
 
--- a/glob/manager_core.py
+++ b/glob/manager_core.py
@@ -324,6 +324,8 @@ def normalize_channel(channel):
         return None
     elif channel.startswith('https://'):
         return channel
+    elif channel.startswith('http://') and get_config()['http_channel_enabled'] == True:
+        return channel
 
     tmp_dict = get_channel_dict()
     channel_url = tmp_dict.get(channel)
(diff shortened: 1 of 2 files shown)


ltdrdata:
It is intentionally configured so that default_cache_update always looks only at the default channel. 
Is this also needed in an airgap environment?

956237586:
yes, i want fully control all network request because network is unstable, hard code core.DEFAULT_CHANNEL make it lost control

956237586:
I even want to add airgap_mode config globally If I have enough time

ltdrdata:
> I even want to add airgap_mode config globally If I have enough time

If that's the case, it might be better to add a separate config item to replace "default_url" instead of using "channel_url."

956237586:
> > I even want to add airgap_mode config globally If I have enough time
> 
> If that's the case, it might be better to add a separate config item to replace "default_url" instead of using "channel_url."

https://github.com/ltdrdata/ComfyUI-Manager/blob/main/glob/manager_core.py#L1588
why using separate config item? default value in config is same as hardcode value

ltdrdata:
> > > I even want to add airgap_mode config globally If I have enough time
> > 
> > 
> > If that's the case, it might be better to add a separate config item to replace "default_url" instead of using "channel_url."
> 
> https://github.com/ltdrdata/ComfyUI-Manager/blob/main/glob/manager_core.py#L1588 why using separate config item? default value in config is same as hardcode value

The default channel serves as the standard reference and is also related to security policies.
The functional differences between the default channel and other channels will gradually increase.

956237586:
> > > > I even want to add airgap_mode config globally If I have enough time
> > > 
> > > 
> > > If that's the case, it might be better to add a separate config item to replace "default_url" instead of using "channel_url."
> > 
> > 
> > https://github.com/ltdrdata/ComfyUI-Manager/blob/main/glob/manager_core.py#L1588 why using separate config item? default value in config is same as hardcode value
> 
> The default channel serves as the standard reference and is also related to security policies. The functional differences between the default channel and other channels will gradually increase.

hard code intentionally for security？

ltdrdata:
More precisely, it refers to the part related to the following code:
https://github.com/ltdrdata/ComfyUI-Manager/blob/ede8279c175e1d11d593bac7a16506577b364999/glob/manager_core.py#L2781

If it's only looking at the `channel_url`, there could be an issue here.
The screening of CNR nodes is performed here. How is access to CNR handled in an airgap environment?
(CNR = comfyregistry)

956237586:
emmm, fully offline env seems need more consideration and modification later
but for my network env, github connection is unstable, but comfyregistry is stable although it slowly

ltdrdata:
For now, since regular users rarely change the `channel_url`, I will merge this.
In the future, I plan to improve the GitHub related code to support reverse proxy.

ltdrdata:
`default_cache_as_channel_url` config option is added.
If you want apply `channel_url` to `default_cache_update`. 
Turn on that option.

PR #1520, opened 2025-02-05
update DB

XieChengYuan:
Since JavaScript code obfuscation was performed in the new repository, the repository address has been updated.

### DIFF
--- a/custom-node-list.json
+++ b/custom-node-list.json
@@ -20162,6 +20162,17 @@
             ],
             "install_type": "unzip",
             "description": "This is a node to convert an image into a CMYK Halftone dot image."
+        },
+        {
+            "author": "lingha",
+            "title": "comfyui_kj",
+            "id": "comfyui_kj",
+            "reference": "https://github.com/XieChengYuan/comfyui_kj",
+            "files": [
+                "https://github.com/XieChengYuan/comfyui_kj"
+            ],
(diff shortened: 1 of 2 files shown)


ltdrdata:
Currently, nodes registered in ComfyUI-Manager are prohibited from using obfuscation due to security policies.

XieChengYuan:
Initially, I thought obfuscation was allowed for JavaScript but not for Python. Now I fully understand the policy and have removed all obfuscated code in compliance with the security requirements. The changes have been pushed to the current branch. Please review the updated code. Thank you for your feedback! 

> Currently, nodes registered in ComfyUI-Manager are prohibited from using obfuscation due to security policies.

XieChengYuan:
Additionally, I have been wondering if there is any relationship between registering nodes through ComfyUI-Manager and registering nodes in the ComfyUI Registry. Do nodes need to be registered in the ComfyUI Registry before they can be registered through ComfyUI-Manager? Or are these two processes independent of each other? There is very little information available online, so I hope to get an answer from you.

ltdrdata:
Registering with ComfyRegistry can be done independently. ComfyRegistry has its own versioning and screening mechanisms. Installing the nightly version is only possible through nodes registered in ComfyUI-Manager.

ltdrdata:
This is added manually.

PR #1530, opened 2025-02-09
Update custom-node-list.json

asdrabael:
Added hunyuan Multi-Lora node for both styles of hunyuan workflows


ltdrdata:
This is added manually.

PR #1532, opened 2025-02-10
Update custom-node-list.json

Amorano:
Updating the old description which seems to be used when the repo is hand installed?

### DIFF
--- a/custom-node-list.json
+++ b/custom-node-list.json
@@ -4033,15 +4033,14 @@
         },
         {
             "author": "amorano",
-            "title": "Jovimetrix Composition Nodes",
+            "title": "Jovimetrix",
             "id": "jovimetrix",
             "reference": "https://github.com/Amorano/Jovimetrix",
             "files": [
                 "https://github.com/Amorano/Jovimetrix"
             ],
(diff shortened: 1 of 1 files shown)


ltdrdata:
Actually, for nodes registered in the registry, the description received from the registry has a higher priority, so updating it here doesn’t have much significance.

Amorano:
> Actually, for nodes registered in the registry, the description received from the registry has a higher priority, so updating it here doesn’t have much significance.

Except..... when I had my repo local it used the description from the file, not the registry -- it also showed my user name as Amorano, not Joviex <--- which means it was pulling from the node_list on github, not the registry.

Hence, I updated it for anyone who has the same "problem".

ltdrdata:
> > Actually, for nodes registered in the registry, the description received from the registry has a higher priority, so updating it here doesn’t have much significance.
> 
> Except..... when I had my repo local it used the description from the file, not the registry -- it also showed my user name as Amorano, not Joviex <--- which means it was pulling from the node_list on github, not the registry.
> 
> Hence, I updated it for anyone who has the same "problem".

I need to take a look at the issue.

PR #1551, opened 2025-02-15
Update custom-node-list.json

mr-pepe69:


mr-pepe69:
i am a bit of a noob. i don't get what to do with the workflow waiting approval.

PR #1563, opened 2025-02-18
Retain workflow versions when serializing node_versions

### LINKED ISSUE #1499: [DevTask] Offer access to the initial nodeVersions metadata after workflow load
After a workflow has been loaded, we should save the initial `nodeVersions` metadata somewhere so that ComfyUI-Manager can utilize it to install exact versions of custom nodes to reproduce a workflow.

christian-byrne:
- Change from writing `node_versions` on outer level of workflow data to `properties` on each individual node, which prevents issues with tracking deleted nodes and linking package names with node names.
- Change from writing in graph serialization, which happens multiple times per second, to writing in `nodeCreated`, as node.properties can be loaded and persisted from workflow json unlike the `extra` field.
- Since it's not necessary to patch serialization with this approach, workflow and local versions can be accessed without additional steps. For example:

```js
const localNodes = app.graph.serialize().nodes
const workflowNodes = JSON.parse(app.extensionManager.workflow.activeWorkflow.content)
  .nodes
```

To test:

```js
const printTable = (nodes) => {
  nodes.forEach((node) => {
    console.table(
      Object.entries(node.properties).map(([key, value]) => ({
        Property: key,
        Value: value
      }))
    )
  })
}

printTable(localNodes)
printTable(workflowNodes)
```

Resolves https://github.com/ltdrdata/ComfyUI-Manager/issues/1499

### DIFF
--- a/js/workflow-metadata.js
+++ b/js/workflow-metadata.js
@@ -47,46 +47,36 @@ class WorkflowMetadataExtension {
     this.comfyCoreVersion = (await api.getSystemStats()).system.comfyui_version;
   }
 
-  setup() {
-    const extension = this;
-    const originalSerialize = LGraph.prototype.serialize;
+  /**
+   * Called when any node is created
+   * @param {LGraphNode} node The newly created node
+   */
+  nodeCreated(node) {
+    try {
+      // nodeData doesn't exist if node is missing or node is frontend only node
+      if (!node?.constructor?.nodeData?.python_module) return;
 
-    LGraph.prototype.serialize = function () {
-      const workflow = originalSerialize.apply(this, arguments);
-      const graph = this;
+      const nodeProperties = (node.properties ??= {});
+      const modules = node.constructor.nodeData.python_module.split(".");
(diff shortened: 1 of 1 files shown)


huchenlei (approved):

huchenlei (commented):
Looking into greater detail, will this PR makes saving the workflow always persisting the original versions of nodes when initially creating the workflow (Even if relevant nodes are removed)?

christian-byrne:
> Looking into greater detail, will this PR makes saving the workflow always persisting the original versions of nodes when initially creating the workflow (Even if relevant nodes are removed)?

Yes, that will happen actually.

Since the workflow format just lists node versions instead of mapping them to nodes, it is hard to distinguish between removed vs missing. I think it would be necessary to (1) query each `node_version`, (2) get the list of custom node names for the `node_version`, (3) cross reference with all nodes in graph, (4) if no matches found, remove; if match found, keep -- as it corresponds to the missing node. 

A similar process would be necessary when updating custom nodes locally in between saves.

PR #1574, opened 2025-02-22
added daves_nodes to custom node list for pull request

GamingDaveUk:
Created a custom node, that is used in some of my workflows and needed to make it so others could install it with ease before I share my workflows. 

Created a fresh install of comfyui, added the node into the file and switched to local database. It did not detect the node in "missing nodes" however when i searched for my node in the custom node list I was able to install it, a restart then had the node working fine.... no idea if it not being detected in missing nodes is simply down to the node being added to my workflow before the custom node method.... but the custom-node-list is working in local db as requested in the docs. hoping i am submitting it correctly! been a long time since i last did pull requests lol

### DIFF
--- a/custom-node-list.json
+++ b/custom-node-list.json
@@ -21545,6 +21545,17 @@
             ],
             "install_type": "unzip",
             "description": "This is a node to convert an image into a CMYK Halftone dot image."
+        },
+        {
+            "author": "GamingDaveUk",
+            "title": "Daves Nodes",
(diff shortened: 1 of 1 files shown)


ltdrdata:
Please remove `__pycache__` dir from the repo.

GamingDaveUk:
> Please remove `__pycache__` dir from the repo.

have removed from daves_nodes repo

PR #1578, opened 2025-02-25
restart action: support running as Python module

bigcat88:
This little PR simply adds a check if the main python executable file has the name `__main__.py` - then the restart is performed via `python -m MODULE_NAME ...`, and not directly via the full path to "__main__.py", which leads to a `relative imports` error, since python modules always do relative imports within themselves.

This will allow us to use ComfyUI-Manager in future versions of Visionatrix, our project is launched by the line:

`python -m visionatrix`

With these changes, the restart works successfully, and these changes are not tied specifically to our project, but are universal, and may be useful to someone else in the future.


bigcat88:
thank you

PR #1587, opened 2025-02-27
add comfyui-finegrain to custom-node-list.json

Laurent2916:
small set of nodes to interact with an API
https://github.com/finegrain-ai/comfyui-finegrain


Laurent2916:
:wave: @ltdrdata, is there something specific I need to do with this PR to get it merged?
It's been stuck in limbo for a couple days now.

ltdrdata:
The dependencies in `requirements.txt` are pointing to a package within the repository file instead of being distributed via PyPI, which has led to a policy-based pending status.

Due to prior cases, a rather strict policy is being applied to this.

Laurent2916:
Got it, I didn't about this constraint. I modified the `requirements.txt` to not depend on dependencies not on PyPI.

PR #1589, opened 2025-02-27
Add new nodes to ComfyUI_pytorch360convert extension

ProGamerGov:


ltdrdata:
You should not update that file manually. It will be updated on my side by script.

PR #1615, opened 2025-03-05
small fix for running as py-module on windows

bigcat88:
In PR I simply changed the conditions for the code, for "python module" escaping is not important, so it can be first to not increase size of code.

I didn't notice then that there is a separate logic for Windows (_I wonder why? - Linux supports paths with `"`_) - and now that we have successfully switched to ComfyUI-Manager! test actions on Windows are starting to fail :(

P.S.: I saw another strange log string: I constantly see the text `'comfyui-frontend-package' dependency were fixed` - since the package name is actually `comfyui_frontend_package` and code checks for `comfyui-frontend-package` - should I create an issue and ask you whether to fix this or not (this is the same PEP 503, where `-` and `_` are equal in names) - or should I just open PRs right away?

### DIFF
--- a/glob/manager_server.py
+++ b/glob/manager_server.py
@@ -1599,11 +1599,11 @@ def restart(self):
     if '--windows-standalone-build' in sys_argv:
         sys_argv.remove('--windows-standalone-build')
 
-    if sys.platform.startswith('win32'):
-        cmds = ['"' + sys.executable + '"', '"' + sys_argv[0] + '"'] + sys_argv[1:]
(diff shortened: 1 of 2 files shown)


ltdrdata:
The issue with `comfyui-frontend-package` and `comfyui_frontend_package` occurred because `pip freeze` and `uv pip freeze` return different results. This issue has been fixed.

ltdrdata:
The dedicated logic for Windows exists because of the following reason:
https://github.com/ltdrdata/ComfyUI-Manager/issues/620
https://github.com/ltdrdata/ComfyUI-Manager/pull/621

PR #1619, opened 2025-03-06
Update Image Position Blend 1.1

S4MUEL-404:
Added Layer Image Masking feature, supporting transparent background images.
Added Layer Image Scaling feature.
Adjusted the display order of input interfaces.


ltdrdata:
This is added already.

PR #1620, opened 2025-03-06
Add ComfyUI-Text-On-Image 1.0 to manager list

S4MUEL-404:
You can display custom text on an image and set the text size, font, position, and shadow.


ltdrdata:
This is addded manually.

PR #1622, opened 2025-03-06
uv detection based on .venv path

nomeforme:
(claude 3.7 generated description based on changes)

Improved UV Detection in Security Check
This PR addresses an inconsistency in how UV is detected between the security check module and the manager utility functions.

Currently, `security_check.security_check()` runs during ComfyUI prestartup before any of the functions from `manager_util.py` are used. The former does not implement checks for uv usage.

Instead of relying on a configurable flag, I've updated both implementations to directly check for the existence of UV in the venv path. This approach is more reliable since UV installations typically don't include pip by default.

The change ensures consistent behavior between the security check process and other manager utilities when detecting and using the appropriate package manager (UV or pip).

### DIFF
--- a/glob/manager_util.py
+++ b/glob/manager_util.py
@@ -22,9 +22,6 @@ cache_lock = threading.Lock()
 comfyui_manager_path = os.path.abspath(os.path.join(os.path.dirname(__file__), '..'))
 cache_dir = os.path.join(comfyui_manager_path, '.cache')  # This path is also updated together in **manager_core.update_user_directory**.
 
-use_uv = False
-
-
 def add_python_path_to_env():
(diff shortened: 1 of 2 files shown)


ltdrdata:
Oh, it's because that is still in the testing phase, so UV is not set as the default. After further testing, if no issues arise, I will make it the default setting.

This isn't just applied to the security scanner—it is being used across all dependency-related areas.

nomeforme:
Thanks for clarifying! Makes sense that this would need application to all dependency-related areas. Ill use this locally until the official default is released. I'm happy to close this PR if you feel its not relevant at this stage.

PR #1634, opened 2025-03-09
Custom Nodes list: Fixed the formating and added my missing nodes and descriptions

if-ai:
Sorry for missing the id's earlier. Thank you for the help.


if-ai:
I can see the nodes now thanks

PR #1657, opened 2025-03-13
set channel=default, mode=cache for git clone

bigcat88:
Related to https://github.com/ltdrdata/ComfyUI-Manager/issues/1651#issuecomment-2720360147

I'm not sure if this is the correct fix, but judging by the code above, at the start of `gitclone_install` function - it is assumed that `channel=default` and `mode=cache`

That's why I set them to be passed to `unified_manager.install_by_id` - from what I understand of the code, that's how it should be (but I could be wrong, I'm still figuring out this idea of ​​different channels)


ltdrdata (approved):

ltdrdata:
Oh... I missed this PR.
This code is reflected by myself.

PR #1680, opened 2025-03-21
ComfyUI-ergouzi-SVDlora

11dogzi:
EGSVD LoRA Tools is a ComfyUI extension for converting and loading LoRA files with quantized svdqnunchaku models. It provides a node that allows you to convert and load regular LoRAs to be compatible with quantized svdqnunchaku models in real-time.


11dogzi:
The official has already developed this feature, but it has been duplicated and will no longer be submitted

PR #1712, opened 2025-04-03
feat: add hal.fun hugging face model downloader

michaelgold:


ltdrdata:
This is added manually.

PR #1726, opened 2025-04-09
Add workflow to publish to PyPi on version bump

christian-byrne:
Mirrors release workflow from https://github.com/Comfy-Org/ComfyUI_frontend/blob/main/.github/workflows/release.yaml, publishing to PyPi on change to pyproject.toml, as with publish to registry workflow.


christian-byrne:
Will be included in https://github.com/Comfy-Org/ComfyUI-Manager/tree/feat/cacheless-v2

PR #1727, opened 2025-04-09
Added HiDream Sampler

benjamin-bertram:
A custom ComfyUI node for generating images using the HiDream AI model. Uses quantization for lower memory usage.

https://github.com/lum3on/comfyui_HiDream-Sampler


ltdrdata:
This node is added manually.

PR #1741, opened 2025-04-13
Add custom node xLegende/ComfyUI-Prompt-Formatter

xLegende:
Custom nodes for formating prompts


ltdrdata:
Please either rename `install.py` or remove the manual pip installation from it.  
pip installations must be managed exclusively by ComfyUI-Manager.

xLegende:
Thanks! I've removed `install.py`

PR #1768, opened 2025-04-24
Add fixed MagicQuill node

brantje:
Fixed the MagicQuill node. 
Planning to add SDXL and FLUX support at a later stage.

### DIFF
--- a/custom-node-list.json
+++ b/custom-node-list.json
@@ -23869,6 +23869,17 @@
             "description": "Adds extra API functionallity and prometheus endpoint"
         },
         {
+            "author": "brantje",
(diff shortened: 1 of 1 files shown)


ltdrdata:
https://github.com/brantje/ComfyUI_MagicQuill/blob/0c7a4ff3a5fefec0e472344f659ed0153a5c78ff/install.py#L17

Do not clone other nodepacks during the install phase.
If the installation of another nodepack is required, clearly state this in the description, and if a required nodepack is missing when imported, display a warning instead.

brantje:
Hi!
I have removed the nodepacks from the install, please let me know if there are more changes required =).

PR #1803, opened 2025-05-08
ComfyUI-DD-Translation and ComfyUI-DD-Nodes plugin registration

Dontdrunk:
ComfyUI-DD-Translation and ComfyUI-DD-Nodes plugin registration


Dontdrunk (commented):
No problem

PR #1879, opened 2025-05-27
Update custom-node-list.json (Add ComfyUI-E-Tier-TextSaver to node list)

e-tier-newbie:
hi there,

this PR adds ComfyUI-E-Tier-TextSaver node for saving cleaned text outputs. It removes unwanted tokens like <pad> and saves to .txt.

thank you! :)

### DIFF
--- a/custom-node-list.json
+++ b/custom-node-list.json
@@ -26731,9 +26731,9 @@
             "id": "e-tier-text-saver",
             "reference": "https://github.com/e-tier-newbie/ComfyUI-E-Tier-TextSaver",
             "files": [
-                "https://github.com/e-tier-newbie/ComfyUI-E-Tier-TextSaver/archive/refs/heads/main.zip"
+                "https://github.com/e-tier-newbie/ComfyUI-E-Tier-TextSaver"
             ],
-            "install_type": "unzip",
+            "install_type": "git-clone",
             "description": "A node for saving cleaned text outputs, useful for LoRA training. Removes unwanted tokens like <pad> and saves to .txt."
         }
     ]
--- a/custom-node-list.json
+++ b/custom-node-list.json
@@ -26724,6 +26724,17 @@
             ],
             "install_type": "unzip",
             "description": "This is a node to convert an image into a CMYK Halftone dot image."
+        },
+        {
+            "author": "e-tier-newbie",
+            "title": "ComfyUI-E-Tier-TextSaver",
+            "id": "e-tier-text-saver",
+            "reference": "https://github.com/e-tier-newbie/ComfyUI-E-Tier-TextSaver",
+            "files": [
+                "https://github.com/e-tier-newbie/ComfyUI-E-Tier-TextSaver/archive/refs/heads/main.zip"
+            ],
+            "install_type": "unzip",
+            "description": "A node for saving cleaned text outputs, useful for LoRA training. Removes unwanted tokens like <pad> and saves to .txt."
         }
     ]
 }


ltdrdata:
https://github.com/e-tier-newbie/ComfyUI-E-Tier-TextSaver/blob/975bf75bcbb3ba3a2cb0c49e3f2ebd7811ed71d9/py/e_tier_text_saver.py#L39

Your node writes files to arbitrary paths. This is a critical vulnerability as it can lead to Remote Code Execution.

1. Please strictly restrict the location where files are created.
2. Ensure that it is not possible to escape the controlled directory scope using relative paths or similar methods.

e-tier-newbie:
Hi there! 🙇‍♂️

Thank you so much for your valuable feedback — I truly appreciate your time and thoughtful suggestions.

Based on your input, I've carefully revised the implementation with a focus on security:

🔐 Rather than hardcoding a fixed output path, I opted to enforce strict path validation.
This approach preserves flexibility for users while ensuring safe file operations.

Below is a detailed breakdown — I hope it addresses your concerns appropriately.
Please feel free to suggest further changes if needed!




📌 Use Case Motivation

This node is intended for use after Florence captioning nodes in workflows for LoRA training.
Users typically need to generate .txt caption files within specific dataset directories.
Hence, allowing controlled output to arbitrary (but safe) paths is essential.


image




🔐 Security Enhancements

To mitigate the risk of arbitrary file writes, the following validations have been added:

    ❌ Access to system directories is blocked (e.g., C:\Windows)

    ❌ Relative path traversal is denied (..)

    ❌ Dangerous characters are filtered (|, ;, $, etc.)

    ❌ Only existing directories are allowed (auto-creation is disabled)

✅ Resulting Behavior

    Users can only save to existing, validated directories

    Unsafe paths are rejected with clear error messages

    Input text is sanitized to strip harmful patterns (e.g., eval, <script>)

    The final output path is shown clearly in the console



This aims to strike the right balance between usability for dataset generation and robust security.

Let me know if you'd like to see any further refinements!

Thank you again 🙏

ltdrdata:
Controlling the write path should not be done using a blacklist approach; instead, it must be based on a whitelist explicitly specified by the user. Furthermore, the management of this whitelist must reside in an area that cannot be modified remotely.

If a blacklist approach is used, any paths overlooked by the developer—either unintentionally or due to lack of awareness—may become attack vectors. For example, an attacker could modify the code of ComfyUI itself or installed custom nodes, create a fake custom node, or weaken security settings.

e-tier-newbie:
🔒 Updated to Use Explicit, Immutable Whitelist for Output Path

Thank you for your feedback. Based on your review, I've restructured the path control mechanism as follows:

    ✅ Replaced dynamic dictionary with an explicit whitelist (WHITELIST_LABELS) defined at the module level.

    ✅ Wrapped the whitelist with MappingProxyType to ensure immutability and prevent any modification at runtime.

    ✅ Validated the resolved path against symbolic link attacks and ensured it stays within the whitelisted scope.

These changes ensure that the output path is strictly constrained to a user-specified and developer-controlled safe area. The whitelist cannot be altered remotely or at runtime, and directory traversal or symlink exploits are not possible under this structure.

Let me know if any further improvements are needed. I appreciate your guidance in strengthening this node’s security.

ltdrdata:
https://github.com/ltdrdata/was-node-suite-comfyui/commit/f86447721c18758d7b3b6ae5769f52344a3d764a

It might be better to let users configure the whitelist file and restrict access to the specified paths, as done in this patch.

The whitelist file should only be accessible offline.

e-tier-newbie:
In this update, the E_TierTextSaver node is refactored and enhanced by moving output directory whitelist configuration to an external YAML file, enabling users to customize allowed paths offline. The existing input sanitization, path validation, and error handling features are preserved and improved for better security and maintainability.


    Introduced loading of output directory whitelist from config/E_Tier_WL.yaml to allow flexible user configuration.

    Dynamically populate the save_to dropdown options based on the YAML whitelist keys.


Inspired by the design in ltdrdata/was-node-suite-comfyui@f864477, this update improves security and flexibility by moving path restrictions out of hardcoded code and into a user-editable offline YAML configuration. This approach helps maintain safer and more maintainable code while giving users control over output locations.

Thank you always for your valuable review and feedback 🙏🙏🙏
If there is anything else needed, please feel free to let me know anytime.

PR #1905, opened 2025-06-05
Add LoRA-Safe TorchCompile node

xmarre:
Adds the LoRA-Safe TorchCompile node to the community list

### DIFF
--- a/custom-node-list.json
+++ b/custom-node-list.json
@@ -1,14 +1,6 @@
 {
     "custom_nodes": [
         {
-          "name": "LoRA-Safe TorchCompile",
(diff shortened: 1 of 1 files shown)


ltdrdata:
Files in the `lora_safe_compile` path must not be placed inside a subdirectory.
The `__init__.py` file must be located at the root of the repo to function correctly when cloned under `custom_nodes/`.

xmarre:
> Files in the `lora_safe_compile` path must not be placed inside a subdirectory. The `__init__.py` file must be located at the root of the repo to function correctly when cloned under `custom_nodes/`.

Fixed package location, ready for re-check.

PR #1915, opened 2025-06-08
[feat] Implement comprehensive batch tracking and OpenAPI-driven data models

christian-byrne:
Four main goals of this PR:

### Improved Task Tracking

Fixed inconsistencies in the progress tracker dialog during installations. Previously, task state wasn't tracked server-side, forcing the frontend to manually track queue state.

- Tasks now have UUIDs and client IDs for proper identification
- Maintain task history and queue server-side with thread-safe operations
- WebSocket notifications when tasks start/complete keep clients synchronized
- Failed tasks are tracked without interrupting the queue
- Tasks can be cancelled by ID
- Frontend receives consistent stdout logs for each task (previously buggy due to manual tracking)

### Remove Legacy Features

- Cleaning up features being replaced by other projects (native model management, subgraphs)
- Removed unsafe routes like git URL downloads and pip installs - as part of core HTTP API, only registry-validated (security scanned) packages should be downloadable
- Legacy features still available in `legacy` sub-package, accessible via `--enable_manager_legacy_ui` CLI flag

### Type-Safe API-Driven Development

- Generated OpenAPI specification for Manager HTTP API
- Both frontend and backend generate types/SDKs from the same spec, enforcing contracts
- Frontend uses TypeScript checking, backend uses Pydantic validation
- OpenAPI validated against actual server in CI/CD
- Auto-generated documentation

### Set Groundwork for Restore Points

- Captures environment state before/after task batches (ComfyUI version, frontend version, custom nodes, python environment)
- State schemas defined in OpenAPI with validation
- HTTP routes for accessing states (later, migration will be possible as part of snapshot/restore feature)

Additional refactoring includes splitting very large files out into more modules and applying ruff/pylint.

## Test plan
- [x] Test task queue operations with client_id tracking
- [x] Verify batch history files are created correctly
- [x] Test OpenAPI model generation with datamodel-codegen
- [x] Verify all REST endpoints work with new models
- [x] Test queue status filtering by client_id
- [x] Verify tasks are partitioned by failed and succeeded


ltdrdata:
Was it intentional to include `/dist` dir?

christian-byrne:
> Was it intentional to include `/dist` dir?

No. Good catch.

Also, this should be a draft PR until changes are finalized and the test plan is complete.

christian-byrne:
## Test: Client ID Filtering :heavy_check_mark: 

Test that task queue operations work with multiple connected clients and each client only sees the task states that they queued.


https://github.com/user-attachments/assets/a8f70857-dacf-4223-a0f5-10f753bfbedb

christian-byrne:
## Test: Batch history record writing :heavy_check_mark: 

Test that after the task queue is empty for 4 seconds consecutively after tasks were in process, a batch history record is written to disk. The batch history record includes a large number of properties representing the state before and after the operations, as well as the operations themselves.

Batch history records are kept for 16 days and are about 25kb per record (json file).


https://github.com/user-attachments/assets/54c6c7c3-303d-41d6-b6d1-7799986bc5c1

christian-byrne:
## Test: Tasks partitioned based on failed and succeeded in the UI :heavy_check_mark: 

Test that failed tasks are put into separate section in the UI



https://github.com/user-attachments/assets/52d5d0ec-d3e0-4673-a3bc-227baf9fd5a4

christian-byrne:
The test plan is now complete. The next step is to package all three feature branches into easily-testable portable release and begin QA testing. This will start tomorrow.

PR #1980, opened 2025-07-08
Add ComfyUI-exLoadout node to custom-node-list

IsItDanOrAi:
Adds the exLoadout custom node suite, which enables spreadsheet-based loadout management in ComfyUI, including model, CLIP, VAE, and parameter switching via Excel spreadsheet rows.

### DIFF
--- a/custom-node-list.json
+++ b/custom-node-list.json
@@ -29330,5 +29330,16 @@
             "install_type": "unzip",
             "description": "This is a node to convert an image into a CMYK Halftone dot image."
         }
+        {
+            "author": "IsItDanOrAi",
+            "title": "exLoadout: Excel-Based Model & Settings Loader",
+            "id": "comfyui-exloadout",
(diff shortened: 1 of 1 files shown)


ltdrdata:
1. Please remove `__pycache__` dir from the repo.
2. For local file access, paths should be sanitized to enforce restrictions.
Please either use a designated directory or allow specifying directories through a whitelist that can only be edited offline.

https://github.com/IsItDanOrAi/ComfyUI-exLoadout/blob/a7c0ae54ca44a3330ed51f9dcec54bab37693a1a/exLoadoutSelector.py#L31

IsItDanOrAi:
Thank you for getting back to me @ltdrdata and for the feedback. I have made the corrections across all of the nodes in exLoadout and deleted __pycache__ dir from the repo. Hopefully the fix should close hole. If there are any other issues, please let me know, and thank you SO SO much for everything you have done for the community.

PR #1984, opened 2025-07-08
Add Comfy Contact Sheet Image Loader to custom nodes

benstaniford:
A custom ComfyUI node that displays up to 64 recent images from a folder as numbered thumbnails in a flexible grid contact sheet format, allowing you to easily select and load one of them for post-processing.

https://github.com/benstaniford/comfy-contact-sheet-image-loader?tab=readme-ov-file

3vlfrsw7spbf1

### DIFF
--- a/custom-node-list.json
+++ b/custom-node-list.json
@@ -29400,6 +29400,17 @@
             ],
             "install_type": "git-clone",
             "description": "A powerful suite of nodes for ComfyUI to dynamically load prompts and images. This pack is especially focused on   batch processing from a directory and includes two main nodes: Advanced Prompt & Image Loader (Multiple) and   Advanced Prompt & Image Loader (single)."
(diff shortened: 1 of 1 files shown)


ltdrdata:
I added this myself to resolve the conflict.

benstaniford:
@ltdrdata Thanks for that!  Tell me, is there any convention that means these PRs need to add nodes at the bottom of the file?  I was thinking to avoid conflicts, I should stick it in the middle next time.  I'm working on a couple of other nodes at the moment so it'd be good to know.

ltdrdata:
It's usually fine to add node pack at the end, but in this case, the issue happened because I was in the middle of an update, and I had merged some PRs before cleaning them up. 

You forked during that state and appended your code afterwards, which caused the conflict.

PR #2013, opened 2025-07-20
Add ComfyUI-Video-Tools to list

Verolelb:
Hi, this adds my new video tools node. It includes an FPS converter. Thanks!


ltdrdata:
1. To make the repository installable via git clone, an `__init__.py` file must be located at the root of the repo.
2. Remove the `__pycache__` directory from the repository and add it to `.gitignore`.

PR #2059, opened 2025-08-04
Update custom-node-list.json

millerlight:
ComfyUI-KVTools are Nodes for Key/Value (KV) storage, loading JSON inline, from a single file or from multiple JSON files from a folder via dynamic UI dropdowns.

### DIFF
--- a/custom-node-list.json
+++ b/custom-node-list.json
@@ -30867,6 +30867,17 @@
             ],
             "install_type": "unzip",
             "description": "This is a node to convert an image into a CMYK Halftone dot image."
+        },
+        {
+            "author": "millerlight",
+            "title": "millerlight json key-value storage",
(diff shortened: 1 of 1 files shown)


ltdrdata:
Allowing arbitrary files from arbitrary paths to be read is dangerous.
Please restrict file access to a limited scope, for example by using a whitelist that can be configured externally from ComfyUI.

https://github.com/millerlight/ComfyUI-KVTools/blob/e7471b9d941a5411ece166afe316676e3afea362/kv_nodes.py#L136

millerlight:
@ltdrdata Thanks for the feedback — I understand the security concerns and will take them into account.
I’m closing this PR for now and will revisit the implementation with proper file access restrictions in place.
Appreciate your review!

PR #2073, opened 2025-08-11
Millerlight patch 1

millerlight:
Thanks for the earlier security feedback.
In v1.0.0 I now strictly enforce a registry-backed whitelist:

    POST /kvtools/peek only serves values for (file_name, key) that exist in kv_registry.json (basename-only, path traversal blocked via abspath().startswith(BASE_DIR)).

    Image serving is scoped to the allowed root (optional extension whitelist), no arbitrary paths.

    UI changes no longer trigger automatic runs.

### DIFF
--- a/custom-node-list.json
+++ b/custom-node-list.json
@@ -31191,6 +31191,16 @@
             ],
             "install_type": "unzip",
             "description": "This is a node to convert an image into a CMYK Halftone dot image."
+        },
+        {
+            "author": "millerlight",
+            "title": "Key-Value KV-Tools for ComfyUI",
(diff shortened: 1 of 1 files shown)


ltdrdata:
It is recommended to use [folder\_paths](https://github.com/comfyanonymous/ComfyUI/blob/master/[folder_paths](https://github.com/comfyanonymous/ComfyUI/blob/master/folder_paths.py#L114).py#L114) instead of `getcwd`.
Also, it is not recommended to place resource files inside the custom node directory.

https://github.com/millerlight/ComfyUI-KVTools/blob/608b468300f59bffc8f94550d08f55ca9335f451/kv_nodes.py#L178

PR #2077, opened 2025-08-12
Add ComfyUI-S4Tool-Text to custom nodes list

S4MUEL-404:
Text rendering and styling nodes for ComfyUI. This extension provides a basic text renderer, multiple font loaders, and a style node that adds stroke, shadow, gradient fill, and opacity control.


ltdrdata:
This contained unnecessary modifications, so I added it separately.

PR #2090, opened 2025-08-19
update extension-node-map

gzeevi-ltx:
update extension-node-map to include RES4LYF's ClownsharKSampler_Beta.
https://github.com/ClownsharkBatwing/RES4LYF


ltdrdata:
The file is maintained by a script. You don't need to PR it.

PR #2133, opened 2025-09-05
Update custom-node-list.json

tony-zn:
add comfyui-zn-pycode custom node


ltdrdata:
Custom nodes that execute code are not allowed due to security concerns. Your node has been marked as UNSAFE and registered in the dev channel.

PR #2153, opened 2025-09-15
Update custom-node-list.json

Dehypnotic:


ltdrdata:
The file path used for saving files must not be arbitrary. If possible, it should be isolated within the output directory under ComfyUI. If you want to support external locations, please restrict them by using a whitelist configuration file that can be set offline, so that only approved paths are allowed for file saving.

https://github.com/Dehypnotic/comfyui-save-mp3/blob/771981ff5d1dea52795218148eefe17672ef88a3/save_audio_mp3.py#L493C26-L493C48

For ther reference.
https://github.com/ltdrdata/was-node-suite-comfyui/blob/ab9d53c71a83a33d9725a48b570a2f74056e04e2/WAS_Node_Suite.py#L366

https://github.com/ltdrdata/was-node-suite-comfyui/blob/ab9d53c71a83a33d9725a48b570a2f74056e04e2/WAS_Node_Suite.py#L10869

Dehypnotic:
I've added whitelist-configuration as described in the readme, instructions in the warning if saving outside output is attempted, and an empty whitelist with instructions in the node directory.

PR #2204, opened 2025-10-09
feat: updated models-list to add support for Qwen Images models

shawnawshk:
updated models-list to add support for Qwen Images models


ltdrdata:
Please don't reformat the file.

shawnawshk:
> Please don't reformat the file.

Gave back the original empty lines

PR #2221, opened 2025-10-14
Add Dehypnotic Save nodes to custom-node-list.json

Dehypnotic:
I requested name, folders and description changed about a week ago after adding two nodes to the existing one. It was merged to main but soon after removed and placed in the legacy-channel tagged as removed. The old record was placed in dev tagged as unsafe with the explanation: "This nodepack contains a node that has a vulnerability allowing write to arbitrary file paths.". 

I have now retested them all and none of them write to arbitrary file paths without having them in the offline whitelist that I was told to create the first time I submitted the Save MP3-node. Isn't this considered safe anymore?

### DIFF
--- a/custom-node-list.json
+++ b/custom-node-list.json
@@ -22783,6 +22783,16 @@
             "install_type": "git-clone",
             "description": "ComfyUI nodes for face based filtering plus image batch merging."
         },
+		{
+            "author": "Dehypnotic",
+            "title": "Dehypnotic Save nodes",
+            "reference": "https://github.com/Dehypnotic/comfyui-dehypnotic-save-nodes",
+            "files": [
+                "https://github.com/Dehypnotic/comfyui-dehypnotic-save-nodes"
+            ],
+            "install_type": "git-clone",
+            "description": "Save toolkit for audio, image, and video: MP3 audio export with VBR/CBR options, multi-format image saving with workflow/thumbnail metadata, and video + frame encoding (MP4/MKV/WEBM/MOV) — all sharing whitelist-safe paths and rich placeholder templating."
+        },
         {
             "author": "feixuetuba",
             "title": "Spleeter",


ltdrdata:
1. The repository (Dehypnotic/comfyui-dehypnotic-save-nodes) for that node pack is no longer publicly accessible. Therefore, it was considered a removed node pack and moved to the legacy channel.

2. During the review of (Dehypnotic/comfyui-save-mp3), it was found that if a relative path such as .. is used in filename_prefix, it could escape the restricted directory scope.

https://github.com/Dehypnotic/comfyui-save-mp3/blob/d01a8260b213b68866b2d86d15b4d1ef93fbcd5c/save_images.py#L451

It is recommended that the whitelist restriction be applied to the final composed path rather than `resolved_main_dir`, and if possible, that the final validation be performed inside `save_single_image`.

Dehypnotic:
Aha, I see. My bad. I have now applied the whitelist restriction to the final composed path in all 3 nodes.

Dehypnotic:
Still not good enough?

ltdrdata:
After validation, it is still being combined with other strings, which breaks integrity again.
https://github.com/Dehypnotic/comfyui-save-mp3/blob/e6b71436b2f386ee963497f7228fb7ccfe1237e6/save_images.py#L473

It is recommended to perform path validation here:
https://github.com/Dehypnotic/comfyui-save-mp3/blob/e6b71436b2f386ee963497f7228fb7ccfe1237e6/save_images.py#L130

Dehypnotic:
I've now made sure validation is performed on the full path just before saving on all 3 nodes. 

Btw, my intention was to delete https://github.com/Dehypnotic/comfyui-save-mp3 and replace it with https://github.com/Dehypnotic/comfyui-dehypnotic-save-nodes that has the same content, so you can delete the old version in the dev-channel. The one in the legacy-channel is the new as in this add request.

Sorry for my problems getting this right. I hope it is satisfactory now.

ltdrdata:
Please check the status of https://github.com/Dehypnotic/comfyui-dehypnotic-save-nodes
The repository appears to be inaccessible.

Dehypnotic:
Oops, I've now switched it to public visibility.

PR #2227, opened 2025-10-16
Add a Save Image (Dir) node to list

OwlvChirotha:
- Extends the official save_image node
- Allows users to save images to any custom directory
- Maintains panel preview functionally via temporary copies
- Auto-cleans temp files on each run

### DIFF
--- a/custom-node-list.json
+++ b/custom-node-list.json
@@ -34835,6 +34835,22 @@
             ],
             "install_type": "unzip",
             "description": "This is a node to convert an image into a CMYK Halftone dot image."
-        }
+        },
+		{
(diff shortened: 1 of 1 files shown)


ltdrdata:
The `custom_path` introduces a security risk by allowing access to arbitrary paths that are not controlled by the user.
Please restrict path creation to only those within the pre-approved directories defined by a whitelist policy.

https://github.com/OwlvChirotha/ComfyUI-Owlv_Nodes/blob/d2416fd7e67c0b371831abc0bc3e7a371fdd0a8b/__init__.py#L48

PR #2235, opened 2025-10-20
Added ComfyUI Queue Manager node

QuietNoise:
An extension supporting more streamlined prompt queue management.


QuietNoise:
Thank you

PR #2291, opened 2025-11-18
Add ComfyUI-RunpodDirect to custom node list

MadiatorLabs:
Adds ComfyUI-RunpodDirect - a server-side model downloader for Runpod deployments that enables direct downloads to pods with multi-connection support (8 parallel connections), eliminating the need to download models locally and re-upload. (works also on normal ComfyUI instalations)

- Repository: https://github.com/MadiatorLabs/ComfyUI-RunpodDirect
- Install type: git-clone
- Version: 1.0.5
- License: GPL v3

image

image

Tested locally with "DB: Local" mode - loads without errors.

### DIFF
--- a/custom-node-list.json
+++ b/custom-node-list.json
@@ -37009,6 +37009,17 @@
             ],
             "install_type": "unzip",
             "description": "This is a node to convert an image into a CMYK Halftone dot image."
+        },
+        {
+            "author": "Madiator2011",
+            "title": "ComfyUI RunpodDirect",
+            "id": "runpoddirect",
+            "reference": "https://github.com/MadiatorLabs/ComfyUI-RunpodDirect",
+            "files": [
+                "https://github.com/MadiatorLabs/ComfyUI-RunpodDirect"
+            ],
+            "install_type": "git-clone",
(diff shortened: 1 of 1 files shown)


ltdrdata:
Please validate the final `output_path` generated through join. A filename can include relative paths or similar values, which may allow escaping the intended restricted directory.

https://github.com/MadiatorLabs/ComfyUI-RunpodDirect/blob/f5e8bc6c6ec9ed35640b24fb96f3039fe6391f09/__init__.py#L50C48-L50C56

kodxana:
> Please validate the final `output_path` generated through join. A filename can include relative paths or similar values, which may allow escaping the intended restricted directory.
> 
> https://github.com/MadiatorLabs/ComfyUI-RunpodDirect/blob/f5e8bc6c6ec9ed35640b24fb96f3039fe6391f09/__init__.py#L50C48-L50C56

Thank you for pointing this out. I have implemented validation in commit:
https://github.com/MadiatorLabs/ComfyUI-RunpodDirect/commit/f7cc02cccb499e0170d8040d1788bf44598e2709
Update also has been pushed to registry.comfy.org
Thank you for reporting this.

PR #2296, opened 2025-11-19
Add custom node sets to the custom-node-list

EricRollei:
Add custom node sets to the custom-node-list


EricRollei:
Closing to split into individual PRs for easier review

PR #2298, opened 2025-11-20
Add Download Tools for ComfyUI

EricRollei:
Added new download tools for ComfyUI with extensive features for media downloading and web scraping.

### DIFF
--- a/custom-node-list.json
+++ b/custom-node-list.json
@@ -35015,7 +35015,35 @@
             ],
             "install_type": "git-clone",
             "description": "A comprehensive collection of advanced image processing nodes for ComfyUI, featuring state-of-the-art denoising, enhancement, and restoration techniques with GPU acceleration and specialized film grain processing."
-        },        
+        },  
+		{
+		    "title": "Download Tools for ComfyUI",
+		    "reference": "https://github.com/EricRollei/Download_Tools",
+		    "files": [
+		        "https://github.com/EricRollei/Download_Tools"
+		    ],
+		    "install_type": "git-clone",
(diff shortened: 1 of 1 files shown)


ltdrdata:
The output directory should be sanitized.
https://github.com/EricRollei/Download_Tools/blob/43a7391bd101f305b86b65865758b96b3af06d01/nodes/gallery_dl_downloader.py#L779
And the same applies to other parts as well.

EricRollei:
I have updated the repository to address the security concerns.

All nodes now use folder_paths.get_output_directory() to sanitize output paths.
Relative paths are now strictly anchored to the ComfyUI output directory.
Session JSON files are now explicitly ignored in [.gitignore](vscode-file://vscode-app/c:/Users/Eric/AppData/Local/Programs/Microsoft%20VS%20Code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) to prevent leaking credentials.
Please let me know if any other changes are required.

ltdrdata:
Your sanitize patch is incorrectly implemented.
Sanitization is not about error handling — it’s about preventing access to dangerous paths.
Especially for file writes, if your node allows writing to arbitrary locations, a remote attacker can craft a manipulated prompt to exploit your node and plant malicious code.

Because escaping via relative paths is easy, the recommended approach is to validate the final path at the point of use by checking, with `os.path.commonpath`, that it remains isolated within the designated directory.

EricRollei:
Thank you Dr Data for your explanation.  I am using Gemini 3 to make my edits.   With it's and your help, I have updated the path sanitization logic to use [os.path.commonpath](vscode-file://vscode-app/c:/Users/Eric/AppData/Local/Programs/Microsoft%20VS%20Code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) as requested. The code now explicitly verifies that the resolved final path is a subdirectory of the ComfyUI output directory before proceeding. If a path attempts to escape the directory, it reverts to the default output folder.

PR #2316, opened 2025-11-27
Add ComfyUI FSL Nodes — Gemini chat & image generation, utilities

fredlef:
This PR adds the “ComfyUI FSL Nodes” package, including:

- FSLGeminiChat
- FSLGeminiGenerateImage (V8/V9)
- FSL Image Memory Node
- Transparent Background Helper
- Manual Alpha Mask Painter
- 8-Way Image Switch and other utilities

Unified category: “FSL Nodes”
Tested via clean Manager-style installation.
Repository: https://github.com/fredlef/Comfyui_FSL_Nodes

### DIFF
--- a/custom-node-list.json
+++ b/custom-node-list.json
@@ -37560,6 +37560,12 @@
             ],
             "install_type": "unzip",
             "description": "This is a node to convert an image into a CMYK Halftone dot image."
-        }		
+        },
+		{
+   	 		"name": "ComfyUI FSL Nodes",
+    		"url": "https://github.com/fredlef/Comfyui_FSL_Nodes",
+    		"branch": "main",
+    		"tags": ["image", "chat", "gemini", "fsl"]
+  		}
     ]
 }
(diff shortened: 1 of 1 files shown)


ltdrdata:
https://github.com/fredlef/Comfyui_FSL_Nodes/blob/18f69b507c24eb7b4854a94bf4540262ea7d397b/fsl_save_and_strip_alpha.py#L30

When saving files, filename manipulation can be used to corrupt files at arbitrary paths, so the path must be sanitized to ensure that only locations within a restricted scope are allowed. 
Please also apply the same fix to other places where filenames are used.

fredlef:
Thanks for the review! I’ve added full path and filename sanitization to all saving operations:

• fsl_save_and_strip_alpha.py
• FSLImageSaverWithMetadata.py
• FSLImageSaverWithMetadataV5.py
• FSLVeoGenerator (video output)

All save locations are now restricted to the ComfyUI output directory, with
blocked traversal, absolute paths removed, and safe filename normalization.

Let me know if you’d like anything else adjusted.

Thanks,
Fred

fredlef:
“I see a merge conflict in custom-node-list.json. I originally added my entry, but it looks like other submissions have since changed the file. Should I rebase and re-add my entry, or will you handle the conflict on your end?”

PR #2324, opened 2025-11-29
added iSeeBetter Custom Node

llikethat:
iSeeBetter Custom Node for upscaling


llikethat:
Added my custom node to the list. Please let me know if any formatting or metadata updates are needed.

PR #2334, opened 2025-12-02
ComfyUI version listing + nightly current fix

benceruleanlu:
## Summary
- parse and sort ComfyUI semver tags descending, expose a capped list of versions (nightly + current describe + top semvers), and derive `latest_tag` from the highest semver
- treat the default-branch HEAD as `nightly` for the current label when it is not exactly on a tag, while preserving exact tags as the current label
- stable updates now look up the latest tag ref, skip if already on that commit, and log when the tag ref cannot be found

### DIFF
--- a/glob/manager_core.py
+++ b/glob/manager_core.py
@@ -3395,10 +3395,23 @@ def get_comfyui_versions(repo=None):
     except Exception:
         exact_tag = ''
 
+    head_is_default = False
+    remote_name = get_remote_name(repo)
+    if remote_name:
+        try:
+            default_head_ref = repo.refs[f'{remote_name}/HEAD']
+            default_commit = default_head_ref.reference.commit
+            head_is_default = repo.head.commit == default_commit
+        except Exception:
+            head_is_default = False
+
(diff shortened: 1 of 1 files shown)


glob/manager_core.py
```diff
@@ -3356,35 +3359,65 @@ async def restore_snapshot(snapshot_path, git_helper_extras=None):
...
 
-    current_tag = repo.git.describe('--tags')
+    normalized_current = normalize_describe(raw_current_tag)
+    has_normalized_tag = normalized_current in semver_tag_set
+    current_tag = normalized_current if has_normalized_tag else raw_current_tag
 
-    if current_tag not in versions:
-        versions = sorted(versions + [current_tag], key=lambda v: repo.git.log('-1', '--format=%ct', v), reverse=True)
-        versions = versions[:4]
+    if has_normalized_tag and normalized_current not in semver_tags:
```
ltdrdata: This is a contradictory condition.

benceruleanlu: good point

benceruleanlu: fixed

PR #2339, opened 2025-12-02
Add faceExtractor node for ComfyUI

llikethat:
Added faceExtractor node for ComfyUI which identifies based on input image reference


llikethat:
Added my custom node to the list. Please let me know if any formatting or metadata updates are needed.

PR #2349, opened 2025-12-05
Changed Main Dialog to match aesthetics and close button location as Original ComfyUI Interface

GroxicTinch:
Tried to not change as much a I could to prevent things breaking and tried using the classes that the original ComfyUI use so if they update it'll change with it.

image

image

image

image


GroxicTinch:
Sorry for close and reopen, I was unaware the pull request updated with my further commits.

PR #2363, opened 2025-12-08
Add new custom node: lora_downloader

huihuihuiz:
This PR adds the custom node plugin "lora_downloader".

GitHub repo:
https://github.com/huihuihuiz/lora_downloader

description: 
A custom node for downloading, organizing, and managing LoRA models directly inside ComfyUI.

### DIFF
--- a/custom-node-list.json
+++ b/custom-node-list.json
@@ -38193,15 +38193,16 @@
             "install_type": "unzip",
             "description": "This is a node to convert an image into a CMYK Halftone dot image."
         },
-       {
+        {
             "author": "huihuihuiz",
             "title": "LoRA Downloader for ComfyUI",
             "id": "lora_downloader",
             "reference": "https://github.com/huihuihuiz/lora_downloader",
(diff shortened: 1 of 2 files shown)


ltdrdata:
Please use something like os.path.commonpath to ensure that any paths constructed from external input remain within an isolated directory. As it stands, there is a vulnerability that allows writing or reading files at arbitrary locations via relative paths and similar techniques.

https://github.com/huihuihuiz/lora_downloader/blob/28b95a58aa33235adaed7af5dde3f9b1a72abbc8/lora_downloader.py#L43C50-L43C59

https://github.com/huihuihuiz/lora_downloader/blob/28b95a58aa33235adaed7af5dde3f9b1a72abbc8/lora_downloader.py#L108

https://github.com/huihuihuiz/lora_downloader/blob/28b95a58aa33235adaed7af5dde3f9b1a72abbc8/lora_downloader.py#L128

huihuihuiz:
Thanks for the review.

All user-controlled paths have been secured using os.path.commonpath via a safe_join helper, ensuring all file access remains within the allowed directories.

PR #2419, opened 2025-12-23
Add TurboDiffusion I2V entry to custom-node-list

InitialXKO:
Added a new entry for TurboDiffusion I2V with details.


ltdrdata:
The `output_path` should be sanitized to prevent saving to an arbitrary path.

https://github.com/anveshane/Comfyui_turbodiffusion/blob/b0900516e6f96bc99ff1c31e9b545b287ee61e65/nodes/video_saver.py#L139

ltdrdata:
The earlier path-traversal concern has since been resolved in upstream commits, so I've added the entry for `comfyui-turbodiffusion` to `custom-node-list.json` directly. This PR is in a conflict state, so I've added it on my side rather than merging.

PR #2430, opened 2025-12-26
Add ComfyUI-FrameIO node to custom-node-list

room3dev:
Add ComfyUI-FrameIO custom node


ltdrdata:
`file_path` should be restricted.
The current path pattern allows writing files to arbitrary paths.

https://github.com/room3dev/ComfyUI-FrameIO/blob/390de1d360b345dd2ffd3c2a3c2bbb754df46e54/save_webp_sequence.py#L76

PR #2438, opened 2025-12-27
A video analysis and scene extraction tool that generates prompts

GraftingRayman:
A video analysis and scene extraction tool that generates prompts

Uses pyscene, moonbase2

### DIFF
--- a/custom-node-list.json
+++ b/custom-node-list.json
@@ -36665,9 +36665,9 @@
             "author": "GraftingRayman",
             "title": "Border Distance Mapper",
             "id": "iGraftingRayman",
-            "reference": "hhttps://github.com/GraftingRayman/Border-Distance-Mapper",
+            "reference": "https://github.com/GraftingRayman/Border-Distance-Mapper",
             "files": [
-                "https://github.com/GraftingRayman/Border-Distance-Mapper"
+                "ttps://github.com/GraftingRayman/Border-Distance-Mapper"
             ],
             "install_type": "git-clone",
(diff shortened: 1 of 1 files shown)


ltdrdata:
`output_folder` should be sanitized.

https://github.com/GraftingRayman/ComfyUI-Scene/blob/f855c806f95b691e9e3dfb17915dde1305e003ec/VideoSceneGeneration.py#L316C50-L316C63

GraftingRayman:
sanitation complete

ltdrdata:
In the `save` and `read`–related endpoints of https://github.com/GraftingRayman/ComfyUI-Scene/blob/main/routes.py, the `filepath` must be strictly restricted to a specific, isolated directory.
The current implementation is highly vulnerable to path traversal attacks.

GraftingRayman:
Does that include reading? This is strictly for reading images from a path a user provides

ltdrdata:
> Does that include reading? This is strictly for reading images from a path a user provides

APIs like read_description mean that arbitrary files can be read, which in turn allows all information on a user’s computer to be exfiltrated. This kind of behavior must be strictly prohibited.

Restricting access to only image file extensions is relatively safer compared to unrestricted file access, but it is still risky. Therefore, it is strongly recommended to limit the range of files that can be read.

A good approach is to use a whitelist mechanism, defined in an offline-only configuration file, to explicitly specify which paths or files are allowed to be accessed.

GraftingRayman:
Changes made

PR #2492, opened 2026-01-12
Changing Description Of My Custom Node In Node List

JasonHoku:


ltdrdata:
This nodepack is added already.

PR #2499, opened 2026-01-15
Add Preview Video Monitor Pro

CrateTools:
Adding Preview Video Monitor Pro - a custom node that turns second monitors into an interactive preview display and player with user controls. Wipe, compare, side by side, zoom, pan, and color tools let you review your work. The generation tracking system keeps all iterations organized and accessible. Snapshot from the monitor viewer with workflow embedded. Use up to 6 physical monitors.

### DIFF
--- a/custom-node-list.json
+++ b/custom-node-list.json
@@ -41499,6 +41499,17 @@
             "install_type": "copy",
             "description": "A node which takes in x, y, width, height, total width, and total height, in order to accurately represent the area of an image which is covered by area-based conditioning."
         },
+		{
+            "author": "CrateTools",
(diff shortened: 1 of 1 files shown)


ltdrdata:
Please ensure that `test_file` can only access restricted/allowed paths by enforcing path containment checks, for example using `os.path.commonpath`.

https://github.com/CrateTools/comfyui-preview-video-monitor/blob/ad247d5b91db788c16111efa452ab6d65704dca4/__init__.py#L2451

CrateTools:
Fixed! Added path containment checks using os.path.commonpath as requested. Thank you!

PR #2502, opened 2026-01-15
Fix undefined access in custom nodes manager conflicts calculation

jpruiz114:
Add null checks for nodeItem, nodesMap, extItem, and nItem before accessing properties to prevent 'Cannot read properties of undefined' error when processing node conflicts.

Fixes error at custom-nodes-manager.js:1318


jpruiz114:
Hey @ltdrdata, I'm tagging you over here for visibility. Thanks.

PR #2527, opened 2026-01-20
Update custom-node-list.json

zavatmotion:


zavatmotion:
Hi @ltdrdata,
Just checking in to see if everything looks alright with this new PR.
I confirmed that the path sanitization fixes you requested are included in this version. If you need any further changes, please let me know.

Thanks!

ltdrdata:
The sanitizer is weak. You should normalize the path using `normpath` first, and then use `commonpath` to restrict the range of paths that can be accessed.

```
def sanitize_path(path_str, allowed_root):
    if not path_str:
        return None

    clean = path_str.strip().strip('"').strip("'")
    logical = os.path.abspath(os.path.normpath(clean))
    allowed = os.path.abspath(allowed_root)

    try:
        if os.path.commonpath([logical, allowed]) == allowed:
            return Path(logical)
    except ValueError:
        pass

    return None
```

zavatmotion:
Thank you for the guidance. I've implemented the sanitizer using normpath + commonpath exactly as you suggested. Since this is a backup node that needs to write to external drives, I use the drive root as the allowed boundary for absolute paths. Please let me know if you'd like any adjustments.

ltdrdata:
https://github.com/zavatmotion/ComfyUI-Workflow-Backup/blob/c7b5d6d63ed3f1d7b0b9dcf8e4ff08dc40767495/nodes.py#L61

Access to the filesystem root must not be allowed arbitrarily. If arbitrary paths need to be supported, they must be strictly controlled via an offline-configurable whitelist of allowed paths.

zavatmotion:
Hi @ltdrdata,
I have implemented the offline whitelist system as requested.

1. The node now checks for an `extra_paths.json` file in its directory.
2. If the file exists, it allows paths listed there.
3. If the file is missing or the path is not whitelisted, it strictly defaults to the ComfyUI output directory (blocking external access).
4. I have added an `extra_paths.example.json` to the repo to guide users.

Ready for review.

zavatmotion:
Thank you for your time and for the security review. I appreciate the guidance, it definitely made the node safer and better. Happy to contribute to the community!

PR #2557, opened 2026-01-29
Add ComfyUI-MouseCoords to custom-node-list

Malkalypse:
Extension to display real-time mouse coordinates in graph space for ComfyUI workflows.

### DIFF
--- a/custom-node-list.json
+++ b/custom-node-list.json
@@ -6491,6 +6491,17 @@
             "description": "Nodes: Run python tools/download_weights.py first to download weights automatically"
         },
         {
+            "author": "Malkalypse",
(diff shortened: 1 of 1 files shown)


Malkalypse:
This is pretty much the simplest possible actually-useful extension I can think of. I also want to use it as a test case to make sure I understand the process of contributing to the custom node list. Once I know it works, I have other projects I also want to contribute.

ltdrdata:
This node pack has already been added. Only the node pack information has been updated.

Malkalypse:
I noticed it is not seeing it appear in the Custom Nodes Manager. I thought I had the formatting correct for custom-node-list.json but perhaps I made an error?

PR #2567, opened 2026-01-31
Add Photoreal Factory custom node

Miragea-Ss:
Adds Photoreal Factory as a custom node for ComfyUI.
No model weights included; users are expected to place models in the standard ComfyUI models directory.


ltdrdata:
1. `output_folder`, `folder_path`, and `filename` are vulnerable to path traversal attacks. They need proper sanitization.
2. The current directory structure cannot be installed via ComfyUI-Manager. The repository should be structured so that `__init__.py` is located at the root of the repo.


https://github.com/Miragea-Ss/photoreal_factory/blob/17db4fc6139ba5c433453af46541b12dda86ec80/photoreal_factory/nodes.py#L86

https://github.com/Miragea-Ss/photoreal_factory/blob/17db4fc6139ba5c433453af46541b12dda86ec80/photoreal_factory/nodes.py#L62

ltdrdata:
The submitted entry uses non-standard schema fields (`name`/`repo`/`install_type: git`) instead of `title`/`reference`/`install_type: git-clone`, and the diff is based on a stale fork that would delete a large number of existing entries. Please rebase against the latest `main` and resubmit with the standard schema.

Closing this PR for now.

PR #2570, opened 2026-02-01
Add SEngine to custom-node-list

sworksteam:
Extension to use Civitai loras easily from ComfyUI


ltdrdata:
I will merge this, but I recommend fixing it since the URL in `pyproject.toml` is different.

https://github.com/sworksteam/SEngine/blob/d4481dce586bb10b4fcdc33f4e9fb40f2090c009/pyproject.toml#L28

PR #2585, opened 2026-02-06
Added comfyui-outline custom node to custom-node-list.json

dennisvink:
- Adds a new custom node that snaps alpha edges, removes stray islands (keeps largest), and draws an outside-only outline.
  - Inputs: image, width (int), color (hex string).
  - Includes README and node registration under custom_nodes/comfyui-outline/


ltdrdata:
I added this nodepack myself because it needed to be fixed.

dennisvink:
Thanks a bunch for fixing it. Now I have a good reference on the exact format for future custom node PR submissions.

PR #2594, opened 2026-02-07
fix: Replaced link in pyproject.toml with correct one

c0dezer019:
- Original link written in pyproject for comfyregistry leads to a malicious site
- Corrected FreezeFrame entry with correct mapping


ltdrdata:
The changes in this PR were applied separately due to a conflict.

PR #2633, opened 2026-02-21
Add Booru Tagger custom node.

nestflow:
A Booru tagger node built on [pysssss's original node](https://github.com/pythongosssss/ComfyUI-WD14-Tagger), with much performance improvements, API migration, and supports for newer models ([Pixai](https://huggingface.co/deepghs/pixai-tagger-v0.9-onnx) and [Camie v2](https://huggingface.co/Camais03/camie-tagger-v2) taggers currently).

### DIFF
--- a/custom-node-list.json
+++ b/custom-node-list.json
@@ -43627,6 +43627,17 @@
             ],
             "install_type": "unzip",
             "description": "This is a node to convert an image into a CMYK Halftone dot image."
+        },
+        {
+            "author": "nestflow",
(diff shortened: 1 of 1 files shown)


nestflow:
Repo link: https://github.com/nestflow/ComfyUI-Booru-Tagger

nestflow:
Hi, is there any issue or part to fix in this node? I know the original node is classic but rather old, and there might be something uncompliant with current code policy. Plus, I believe many users will be benefited from the newer models and better performance.

PR #2708, opened 2026-03-18
Add ComfyUI-FileExplorer to custom node list

Vltja:
## Summary
Adds [ComfyUI-FileExplorer](https://github.com/Vltja/ComfyUI-FileExplorer) to the custom node list.

A web-based file explorer for ComfyUI that allows browsing, uploading, copying, moving, renaming and deleting files/folders directly from the browser — no Samba or docker cp needed.

## Node Info
- **Author:** Vltja
- **Install type:** git-clone
- **Compatible:** ComfyUI 0.15.0+
- **No workflow nodes** — pure browser-based utility tool

## Features
- Browse the entire ComfyUI directory from the browser
- Upload files (click or drag & drop)
- Create, rename, copy, move, and delete files/folders
- Breadcrumb navigation with back/forward/up
- Sort by name, size, type, or modified date
- Download files directly
- Right-click context menu
- Path traversal protection

## Test plan
- [x] Node installs via git-clone
- [x] Folder icon appears in action bar
- [x] File explorer panel opens and works
- [x] Navigate, upload, delete, copy, move operations work

Screenshot: https://github.com/Vltja/ComfyUI-FileExplorer#readme

### DIFF
--- a/custom-node-list.json
+++ b/custom-node-list.json
@@ -21,7 +21,9 @@
             ],
             "install_type": "git-clone",
             "description": "This node pack offers various detector nodes and detailer nodes that allow you to configure a workflow that automatically enhances facial details. And provide iterative upscaler.\nNOTE: To use the UltralyticsDetectorProvider, you must install the 'ComfyUI Impact Subpack' separately.",
-            "preemptions":["SAMLoader"]
+            "preemptions": [
+                "SAMLoader"
+            ]
         },
         {
             "author": "Dr.Lt.Data",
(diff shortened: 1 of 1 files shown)


ltdrdata:
Hi @Vltja — thanks for the PR! I have to hold `ComfyUI-FileExplorer` — as written it's a security risk on a default ComfyUI install.

It exposes 7 **unauthenticated** web routes (`/file_explorer/list|upload|download|delete|mkdir|move|copy`) performing full filesystem CRUD. The only containment, `_resolve_path()`, uses `os.path.normpath` + `str.startswith(BASE_PATH)`, which is bypassable two ways:
- `startswith(BASE_PATH)` has no trailing separator, so a sibling path like `<BASE_PATH>-evil` passes;
- `normpath` doesn't resolve symlinks, so a symlink under the root escapes it.

These routes are reachable cross-origin (a webpage the user visits can call them against `127.0.0.1`), so this allows arbitrary file read (`/download`) and write/delete/move outside the intended root. Please gate with `realpath` + `os.path.commonpath([base, target]) == base` (not `startswith`), and ideally require auth on these routes.

(Also, please rebase onto the current `custom-node-list.json` — the branch is far behind and its diff would revert many unrelated entries.)

Happy to re-review once that's in. Thanks!

PR #2758, opened 2026-04-04
Fix CRT-Nodes GitHub reference (plugcrypt → PGCRT)

PGCRT:
The CRT-Nodes repository was transferred to a new GitHub account. Update the two URLs still pointing to the old `plugcrypt` username to `PGCRT`.


PGCRT:
Superseded by a combined PR that also adds the missing CRT nodes to the default channel.

PR #2765, opened 2026-04-05
Add artyclaw-comfy nodes

artyclaw:
Adding my custom node pack for prompt logic and RNG


ltdrdata:
DebugTextSaver can write files to arbitrary paths, which is extremely dangerous.

artyclaw:
> DebugTextSaver can write files to arbitrary paths, which is extremely dangerous.

Thank you for the catch. I updated the DebugTextSaver node: "Restricted file writing to the ComfyUI output directory to prevent arbitrary path access."

PR #2791, opened 2026-04-12
Remove my repository entries from dev github stats

silveroxides:
Removed several GitHub repository entries from my nodes as they seem to interfer with users finding them in manager. These nodes have not been experimental in a long time.
Only explanation I can see is that they are still registered in dev for github stats.


ltdrdata:
Those stats items are updated automatically by the scan. Do you want all of those nodepacks removed from the dev channel, or do you only want the stats removed?

silveroxides:
@ltdrdata removed from dev channel. I found nowhere that was possible to do via registry site and that site is so jank already

PR #2798, opened 2026-04-15
Add ComfyUI-OpenPose-Studio and ComfyUI-Ultralytics-Studio to custom-node-list.json

andreszs:
This PR adds two public custom node repositories by andreszs to the Manager database: ComfyUI-OpenPose-Studio and ComfyUI-Ultralytics-Studio. The change only updates the root custom-node-list.json and keeps the JSON valid.

### DIFF
--- a/custom-node-list.json
+++ b/custom-node-list.json
@@ -46255,6 +46255,26 @@
             ],
             "install_type": "unzip",
             "description": "This is a node to convert an image into a CMYK Halftone dot image."
+        },
+        {
+            "author": "andreszs",
+            "title": "ComfyUI-OpenPose-Studio",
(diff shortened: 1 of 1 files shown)


andreszs:
**[OpenPose Studio](https://github.com/andreszs/ComfyUI-OpenPose-Studio)** is a modern pose editor for ComfyUI focused on creating, editing, saving, and reusing OpenPose poses more comfortably than the old editor. It supports pose collections, JSON import/export, background images, and compatibility with legacy pose JSON, so existing workflows can still be reused.

image


**[Ultralytics Studio](https://github.com/andreszs/ComfyUI-Ultralytics-Studio)** is the same idea, but for Ultralytics pose format instead of OpenPose, using the COCO-17 skeleton.

image

PR #2799, opened 2026-04-15
Add ComfyUI-Lora-Pipeline to custom-node-list.json

andreszs:
This PR adds the public custom node repository ComfyUI-Lora-Pipeline by andreszs to the Manager database. The change only updates the root custom-node-list.json and keeps the JSON valid.

### DIFF
--- a/custom-node-list.json
+++ b/custom-node-list.json
@@ -46255,6 +46255,16 @@
             ],
             "install_type": "unzip",
             "description": "This is a node to convert an image into a CMYK Halftone dot image."
+        },
+        {
+            "author": "andreszs",
+            "title": "ComfyUI-Lora-Pipeline",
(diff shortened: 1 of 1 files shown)


andreszs:
**[ComfyUI-Lora-Pipeline](https://github.com/andreszs/ComfyUI-Lora-Pipeline)** provides nodes for building area-based LoRA and conditioning workflows more easily inside ComfyUI. The goal is to simplify multi-region setups where different parts of the image can use different prompts, LoRAs, and strengths without wiring everything manually every time.

It is especially useful for workflows that combine regional prompting with character/control layouts, since it helps keep those complex graphs more organized and reusable. [View full working example workflow.](https://www.andreszsogon.com/building-a-multi-character-comfyui-workflow-with-area-conditioning-openpose-control-and-style-layering/)

image

PR #2806, opened 2026-04-17
Add ComfyUI-SBTools

Amatsukast:
Adding ComfyUI-SBTools - workflow toolkit with background removal (BiRefNet), prompt/image variable system for batch generation, and chroma key tools.

### DIFF
--- a/custom-node-list.json
+++ b/custom-node-list.json
@@ -46338,6 +46338,17 @@
             ],
             "install_type": "unzip",
             "description": "This is a node to convert an image into a CMYK Halftone dot image."
+        },
+        {
+            "author": "Amatsukast",
+            "title": "ComfyUI-SBTools",
+            "id": "sbtools",
+            "reference": "https://github.com/Amatsukast/ComfyUI-SBTools",
+            "files": [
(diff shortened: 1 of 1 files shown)


andreszs:
What'd be the meaning of the _seed_ value in this node?

image

Amatsukast:
> What'd be the meaning of the _seed_ value in this node?
> 
> image

Thanks for reviewing my PR!

The seed value specifically controls the random selection for any connected Prompt Variable nodes that are set to 'Randomize' mode.

It is exposed centrally on this node for two main reasons:
1. Central Control & Symmetry: It allows users to control all 'Random' variables with a single seed, rather than managing individual seeds on multiple variable nodes. This mirrors how the index centrally controls all 'Sequential' variables.

2. Cache Management & Reproducibility: In ComfyUI, if a node's inputs do not change, the cached result is used. By exposing the seed as a widget, users can change its value (via the built-in 'control after generate') to ensure the random variables update in each generation, even when the index is fixed. It also provides a way to reproduce specific random combinations.

I understand that placing the seed widget directly below the index might make it look like it controls the randomization of the index itself. I plan to rename or reposition it in a future update to make its distinct purpose clearer.

PR #2839, opened 2026-04-29
Add ComfyUI-PipelineBarrier to custom node list

brosequist:
## New Node: ComfyUI-PipelineBarrier

**Repository:** https://github.com/brosequist/ComfyUI-PipelineBarrier

**Description:** A single node that flushes PyTorch's GPU memory cache and runs garbage collection between pipeline stages, preventing OOM errors in multi-stage workflows such as two-pass video generation.

### What it does

Calls `soft_empty_cache(force=True)`, `gc.collect()`, `torch.cuda.empty_cache()`, and `torch.cuda.synchronize()` on every visible GPU. Accepts a LATENT input and returns it unchanged — it is a pure synchronization point with no effect on tensor values.

### Why it's useful

Long multi-stage workflows (e.g. WanVideo two-pass I2V with BlockSwap) can OOM at the start of a second stage even after a model has been offloaded, because PyTorch's allocator caches freed blocks rather than returning them to the OS immediately. Placing this node between stages ensures caches are flushed before the next stage begins.

### Checklist

- [x] Node is publicly available on GitHub
- [x] No pip dependencies beyond what ComfyUI already installs
- [x] `NODE_CLASS_MAPPINGS` and `NODE_DISPLAY_NAME_MAPPINGS` exported from `__init__.py`

### DIFF
--- a/custom-node-list.json
+++ b/custom-node-list.json
@@ -1,6 +1,17 @@
 {
     "custom_nodes": [
         {
+            "author": "brosequist",
+            "title": "ComfyUI-PipelineBarrier",
+            "id": "comfyui-pipeline-barrier",
+            "reference": "https://github.com/brosequist/ComfyUI-PipelineBarrier",
+            "files": [
+                "https://github.com/brosequist/ComfyUI-PipelineBarrier"
+            ],
+            "install_type": "git-clone",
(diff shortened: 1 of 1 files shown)


ltdrdata:
Due to merge conflicts, I've added the entry for `ComfyUI-PipelineBarrier` to `custom-node-list.json` directly, using the title and description from this PR.

Closing this PR — the node is now registered.

PR #2851, opened 2026-05-04
Allow disabling prestartup auto pip install/upgrade

caniko:
## Summary

Adds a `dependency_management` config key (with a
`COMFYUI_MANAGER_DEPENDENCY_MANAGEMENT` env override) to opt out of every
prestartup path that mutates the Python interpreter, so Manager can be
used on top of an externally-managed env without screaming subprocess
errors.

## Why

When ComfyUI-Manager runs on an interpreter the user doesn't own — Nix,
Guix, distro Python, locked-down corporate images — every startup ends
the same way:

```
error: The interpreter at /nix/store/.../python3-3.13-env is externally
managed, [...]
[ComfyUI-Manager] Failed to restore comfyui-frontend-package
Command '[..., '-m', 'uv', 'pip', 'install',
'comfyui-frontend-package==1.42.15']' returned non-zero exit status 2.
```

The env already has the right packages pinned by whoever produced it;
Manager doesn't need to reinstall them, and on a read-only store it
can't. Today there's no clean way to tell it to stand down — it tries
`pip` → `python -m uv` → standalone `uv` and only gives up after all
three fail, which produces a noisy startup and a confused user.

## What changes

When `dependency_management = off` (or
`COMFYUI_MANAGER_DEPENDENCY_MANAGEMENT=off`):

- `PIPFixer.fix_broken()` returns immediately. No torch rollback, no
  opencv alignment, no `comfy` uninstall, no `comfyui-frontend-package`
  reinstall.
- The unified dependency resolver block in `prestartup_script.py` is
  skipped before it reaches `uv pip install`.
- Per-node `pip install -r requirements.txt` in
  `execute_lazy_install_script` is skipped.

Read-only operations (`pip list`, `pip freeze`, `pip show`) still run,
so the UI inventory, `cm-cli` listings, and `security_check` keep
working. User-initiated installs via `cm-cli install` and the UI install
buttons are deliberately *not* gated — those are explicit actions and
shouldn't be silently swallowed.

The default is `on`, which preserves today's behavior bit-for-bit. No
existing user is affected unless they flip the switch.

## Configuration

`config.ini`:

    [default]
    dependency_management = off

Or, without touching the config:

    COMFYUI_MANAGER_DEPENDENCY_MANAGEMENT=off comfyui --enable-manager

The env var wins over the config key. Accepted "off" values are `off`,
`false`, `0`, `no`, `disabled` (case-insensitive); anything else is
treated as on.

## Test plan

- [x] Default (key/env unset) — startup behavior matches main
- [x] `dependency_management = off` — ComfyUI starts with
  `--enable-manager` on a Nix-managed env without subprocess errors
- [x] `COMFYUI_MANAGER_DEPENDENCY_MANAGEMENT=off` overrides
  `dependency_management = on`
- [x] `pip list`/`freeze`/`show` paths still work (UI inventory,
  security check)
- [x] `cm-cli install <node>` still installs (CLI not gated)

## Notes

Branched off `manager-v4` (4.2.1). +46/-1 across two files:
`comfyui_manager/common/manager_util.py` and
`comfyui_manager/prestartup_script.py`.

### DIFF
--- a/comfyui_manager/common/manager_util.py
+++ b/comfyui_manager/common/manager_util.py
@@ -28,6 +28,17 @@ use_uv = False
 use_unified_resolver = False
 bypass_ssl = False
 
+# When False, ComfyUI-Manager will not attempt to install, upgrade, or
+# uninstall Python packages on its own (prestartup auto-healing, the unified
+# dependency resolver, and per-node `pip install -r requirements.txt`). This
+# is intended for distro-packaged or otherwise externally-managed Python
+# environments (e.g. Nix, Guix, system package managers, locked-down
+# corporate images) where the interpreter is read-only.
+#
+# Read-only pip operations (`list`, `freeze`, `show`) still run so the UI,
+# inventory, and security checks keep working.
+dependency_management_enabled = True
+
 def is_manager_pip_package():
     return not os.path.exists(os.path.join(comfyui_manager_path, '..', 'custom_nodes'))
 
@@ -412,6 +423,10 @@ class PIPFixer:
         self.manager_files_path = manager_files_path
 
     def fix_broken(self):
+        if not dependency_management_enabled:
+            logging.info("[ComfyUI-Manager] dependency management disabled; skipping PIPFixer.fix_broken()")
+            return
+
         new_pip_versions = get_installed_packages(True)
 
         # remove `comfy` python package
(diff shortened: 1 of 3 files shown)


dwt:
Friendly ping @ maintainers - any reason to not take this in? Do you need anything else to get this merged?

PR #2912, opened 2026-05-24
Add ComfyUI-PixlStash to custom-node-list.json

lindkvis:
* Loads images and metadata from a PixlStash database via the REST API.
* Saves images and metadata back to the database via the REST API.
* Stores tokens and URL in ComfyUI settings
* Nodes for semantic search and likeness search
* Node: https://github.com/Pikselkroken/ComfyUI-PixlStash/
* PixlStash: https://github.com/Pikselkroken/pixlstash/


lindkvis:
I was made aware that the node pack was flagged due to the token secrets being injected into Javascript instead of Python in 1.1.0 -> 1.1.2.

I pushed a new 1.1.3 that addresses this.

lindkvis:
Those changes did not help so I got rid of all environment variable reading instead... so 1.1.5 should not be flagged.

ltdrdata:
1. Write-side path traversal — `filename_prefix` is a free STRING input on `PixlStashPictureSaver` (`OUTPUT_NODE=True`) and flows unmodified into the filename (`f"{filename_prefix}_..."`), then `os.path.join(temp_dir, fname)` + `open(path, "wb")` with no `os.path.basename()`, no `commonpath` containment, no `..` rejection. A prefix like `../../../etc/cron.d/x` escapes `temp_dir`, allowing arbitrary-location file writes.
   - https://github.com/Pikselkroken/ComfyUI-PixlStash/blob/1afa2fe42304419982f9f5bb3e2079d98a2d509a/nodes/picture_saver.py#L54
   - https://github.com/Pikselkroken/ComfyUI-PixlStash/blob/1afa2fe42304419982f9f5bb3e2079d98a2d509a/nodes/picture_saver.py#L139
   - https://github.com/Pikselkroken/ComfyUI-PixlStash/blob/1afa2fe42304419982f9f5bb3e2079d98a2d509a/nodes/picture_saver.py#L155-L158

Please sanitize `filename_prefix` with `os.path.basename()` (or reject any `..` / `/` / `\` characters) before forming the filename.

lindkvis:
Thanks!

I used folder_paths.get_save_image_path() and added a couple of other security fixes.

ltdrdata:
Due to merge conflicts, I've added the entry for `ComfyUI-PixlStash` to `custom-node-list.json` directly, using the title and description from this PR.

PR #2918, opened 2026-05-26
Fix: Addressed issues for ComfyUI-ACES-EXR-Toolkit

azhagurajpandians:
This PR replaces my previous closed PR. I have completely addressed the security issues flagged by the scanner in `ComfyUI-ACES-EXR-Toolkit`.

**Security Fixes Implemented:**
- Removed `imageio` and `imageio[freeimage]` from dependencies to prevent any dynamic/unverified C-binary downloads.
- Removed custom API endpoint `/aces-exr/listdir` and `js/exr_browser.js` which were flagged for arbitrary directory browsing/traversal.
- Secured `/nodex_hdr/view` endpoint by strictly using `os.path.basename()` to restrict access only to the temporary directory.
- Removed all usage of `ctypes`.
- Completely transitioned to using ComfyUI's native OpenCV dependency (`cv2`) for EXR float32 support.

The node is now secure and fully compliant with the registry's safety guidelines. Let me know if there is anything else I need to address!

### DIFF
--- a/custom-node-list.json
+++ b/custom-node-list.json
@@ -52834,9 +52834,10 @@
                 "https://civitai.com/api/download/models/158997"
             ],
             "install_type": "unzip",
-            "description": "This is a node to convert an image into a CMYK Halftone dot image.",
-        "description": "ComfyUI custom nodes for one-sample-per-run iteration across image, video, and text sources.",
-        "author": "azhagurajpandians",
+            "description": "This is a node to convert an image into a CMYK Halftone dot image."
+        },
+        {
+            "author": "azhagurajpandians",
             "title": "ComfyUI ACES EXR Toolkit",
             "reference": "https://github.com/azhagurajpandians/ComfyUI-ACES-EXR-Toolkit",
             "files": [
--- a/custom-node-list.json
+++ b/custom-node-list.json
@@ -52834,7 +52834,17 @@
                 "https://civitai.com/api/download/models/158997"
             ],
             "install_type": "unzip",
-            "description": "This is a node to convert an image into a CMYK Halftone dot image."
+            "description": "This is a node to convert an image into a CMYK Halftone dot image.",
+        "description": "ComfyUI custom nodes for one-sample-per-run iteration across image, video, and text sources.",
+        "author": "azhagurajpandians",
+            "title": "ComfyUI ACES EXR Toolkit",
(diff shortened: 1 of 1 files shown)


ltdrdata:
Hi @azhagurajpandians — thanks for resubmitting! I re-reviewed `ComfyUI-ACES-EXR-Toolkit` at the current HEAD, and the path-handling issues from the earlier review are still present, so I'm holding this until they're addressed:

1. **`ACESSaveEXR`**: `output_dir` + `filename` aren't confined to the ComfyUI output directory — absolute or `..` paths let a workflow write an `.exr` anywhere on the host. Please resolve the final path and require it to stay under the output dir (`realpath` + `commonpath` containment), rejecting anything that escapes.
2. **`ACESLoadEXRFromPath`** and **`VispyEXRViewer`**: `exr_path` is read from any absolute path (only an existence check). `VispyEXRViewer` additionally stages the read bytes into a temp file that's served over the node's HTTP route, so this is a read-and-return path. Please apply the same containment (confine reads to the input/output/allowed dirs).

These are workflow-reachable (a shared/malicious workflow can set the string widgets), so containment is the fix rather than input sanitization alone.

Happy to re-review once these are in. Thanks!

azhagurajpandians:
"Hi @ltdrdata —"Thanks for the detailed review! For ACESLoadEXRFromPath, this node is explicitly designed for VFX/render-farm pipelines where EXR sequences live on any drive or network path. Data only flows as a tensor within ComfyUI — it is never served over HTTP or exfiltrated. VispyEXRViewer (which stages bytes over HTTP) and ACESSaveEXR both retain full realpath + commonpath containment. Happy to discuss if there's a better approach!"

ltdrdata:
Re-reviewed at the current HEAD. The two side-effecting sinks are now contained:

- `ACESSaveEXR` resolves the output path and enforces `realpath` + `commonpath` containment under the output dir before writing, and strips directory separators from the filename.
- `VispyEXRViewer`'s disk-read path applies the same containment (input/output dirs) before staging, and the `/nodex_hdr/view` route serves only a basename from the temp dir.

`ACESLoadEXRFromPath` still reads an arbitrary path, but with the viewer's read path now confined there's no HTTP return channel for it, so it's a contained read-into-tensor. Registered to the manager node list. Thanks for the fix.

PR #2919, opened 2026-05-26
Add ComfyUI-TrixLoader custom node

trx7111:
Please add ComfyUI-TrixLoader to the custom node list. It is an all-in-one image loader with interactive crop editor, mask editor, and color grading controls.

### DIFF
--- a/custom-node-list.json
+++ b/custom-node-list.json
@@ -52595,6 +52595,16 @@
             ],
             "install_type": "git-clone",
             "description": "ComfyUI custom nodes for one-sample-per-run iteration across image, video, and text sources."
+        },
+        {
+            "author": "trx7111",
+            "title": "ComfyUI-TrixLoader",
+            "reference": "https://github.com/trx7111/ComfyUI-TrixLoader",
+            "files": [
+                "https://github.com/trx7111/ComfyUI-TrixLoader"
+            ],
+            "install_type": "git-clone",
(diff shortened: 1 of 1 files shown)


ltdrdata:
Thanks for the submission. I reviewed the code before registering and have to hold this — there's an arbitrary file write in a web route that needs fixing first.

**Arbitrary file write → RCE.** `api_save_image_with_mask` (route `POST /trix/save_image_with_mask`, `trix_loader_nodes.py:3794`). The primary save path is confined via `commonpath` against `upload_dir`, but the optional `save_every_step` branch (`:3894`) bypasses it entirely:

```python
custom_path = post.get("save_every_step_path", "").strip()   # untrusted
if not os.path.isabs(custom_path):
    target_dir = os.path.abspath(os.path.join(comfy_root, custom_path))  # no '..' rejection
else:
    target_dir = os.path.abspath(custom_path)                # no containment at all
os.makedirs(target_dir)
target_file = os.path.join(target_dir, filename)             # filename also attacker-controlled
shutil.copy2(filepath, target_file)
```

An absolute `save_every_step_path` (or a relative one with `../`) plus the attacker-controlled `filename` writes the uploaded image bytes to any path the server user can write — e.g. `custom_nodes/<x>/__init__.py`, giving remote code execution. The route is unauthenticated and reachable cross-origin via a CSRF-submitted request; the write happens server-side, so CORS doesn't mitigate it.

Please confine `target_dir`/`target_file` to the intended upload/output root (the same containment you already apply to the primary path), then re-request review and I'll re-scan.

PR #2926, opened 2026-05-28
Add ComfyUI Text Preset Loader node

INuBq8:
this is a node to save and organise preset of texts. I always have certain prompts preset I use either for styles, characters, clothes etc. and since I am lazy figured I create this node which allow me to save the texts presets in json and load them at will.


ltdrdata:
Due to merge conflicts, I've added the entry for `ComfyUI-text-preset-loader` to `custom-node-list.json` directly, using the title and description from this PR.

PR #2946, opened 2026-06-03
Add ComfyUI-WanAnimatePlus

wuwukaka:
Add ComfyUI-WanAnimatePlus

### DIFF
--- a/custom-node-list.json
+++ b/custom-node-list.json
@@ -52855,6 +52855,16 @@
             ],
             "install_type": "unzip",
             "description": "This is a node to convert an image into a CMYK Halftone dot image."
+        },
+        {
+            "author": "wuwukaka",
+            "title": "ComfyUI-WanAnimatePlus",
+            "reference": "https://github.com/wuwukaka/ComfyUI-WanAnimatePlus",
+            "files": [
+                "https://github.com/wuwukaka/ComfyUI-WanAnimatePlus"
+            ],
+            "install_type": "git-clone",
+            "description": "Seamless video connection and multi-reference image support for Wan Animate."
         }
     ]
 }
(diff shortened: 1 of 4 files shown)


ltdrdata:
Thanks for the submission. I reviewed the code before registering and have to hold this — there's an unconfined file-write path that needs fixing first.

**Path-traversal write.** `WanVideoImageToVideoMultiTalk` (`multitalk/nodes.py:412`) and `WanVideoImageToVideoSkyreelsv3_audio` (`:495`) both expose a free STRING `output_path` input that flows into a write sink with no containment:

```python
output_path = ("STRING", {"default": ""})            # node widget input
...
if output_path:
    output_path = os.path.join(output_path, f"{timestamp}_{mode}_output")
    os.makedirs(output_path, exist_ok=True)          # no containment
```

The value is then consumed in `multitalk/multitalk_loop.py:503`:

```python
im.save(os.path.join(output_path, f"frame_{img_counter:05d}.png"))
```

There's no `basename`, no `realpath`+`commonpath` confinement, no `..` rejection, and no `folder_paths` root. An absolute or `../`-laden `output_path` escapes the ComfyUI output directory and creates directories / writes frame PNGs at an attacker-chosen location (arbitrary-directory write and file overwrite).

This is remotely reachable even though it's a node widget rather than a web route: `POST /prompt` accepts arbitrary workflow graphs cross-origin with no auth/CSRF token, and the widget string is submitted as part of that graph, so a malicious page the victim visits can drive this write. The write is server-side, so CORS doesn't mitigate it.

Please confine `output_path` to the ComfyUI output root (e.g. resolve under `folder_paths.get_output_directory()` and reject paths that escape it), then re-request review and I'll re-scan.

PR #2949, opened 2026-06-04
Add Frog Node Pack

RabbitThatIsPink:
## New node addition

- **Author:** RabbitThatIsPink
- **Title:** Frog Node Pack
- **Repository:** https://github.com/RabbitThatIsPink/FrogNodePack
- **Install type:** git-clone
- **Description:** A custom node pack for ComfyUI focused on keeping workflows clean and consolidated.

### DIFF
--- a/custom-node-list.json
+++ b/custom-node-list.json
@@ -52915,6 +52915,16 @@
             ],
             "install_type": "unzip",
             "description": "This is a node to convert an image into a CMYK Halftone dot image."
+        },
+        {
+            "author": "RabbitThatIsPink",
+            "title": "Frog Node Pack",
+            "reference": "https://github.com/RabbitThatIsPink/FrogNodePack",
+            "files": [
+                "https://github.com/RabbitThatIsPink/FrogNodePack"
+            ],
+            "install_type": "git-clone",
+            "description": "A custom node pack for ComfyUI focused on keeping workflows clean and consolidated."
         }
     ]
 }


ltdrdata:
Thanks for the submission. I reviewed the code before registering and have to hold this — there's an unconfined file-write path that needs fixing first.

**Path-traversal write.** `RibbitySaveA1111` (`ribbity_save.py:908`) and `RibbitySaveHashEmbed` (`:1024`) both take a free STRING `output_path` that flows into `_resolve_output_dir` (`:765`) with no containment:

```python
custom = (output_path or "").strip()
if custom:
    resolved = os.path.expanduser(custom)
    if not os.path.isabs(resolved):
        resolved = os.path.join(default_output, resolved)  # no '..' rejection
    os.makedirs(resolved, exist_ok=True)                   # absolute path used verbatim
```

`resolved` is then the directory that `_save_frames` writes the image PNGs into. An absolute `output_path` is honored verbatim, and a relative one with `../` escapes `default_output` — no `basename`, no `realpath`+`commonpath` confinement, no `..` rejection. A workflow can create directories and write image bytes at an attacker-chosen location (arbitrary-directory write / file overwrite).

This is remotely reachable even though it's a node widget rather than a web route: `POST /prompt` accepts arbitrary workflow graphs cross-origin with no auth/CSRF token, and the widget string is submitted as part of that graph. The write is server-side, so CORS doesn't mitigate it.

Please confine `output_path` to the ComfyUI output root (resolve under `folder_paths.get_output_directory()` and reject paths that escape it), then re-request review and I'll re-scan.

RabbitThatIsPink:
Changes should be up now :)

ltdrdata:
Thanks — I confirmed the `output_path` fix in `RibbitySaveA1111` / `RibbitySaveHashEmbed` (the `realpath`+`commonpath` check in `_resolve_output_dir` is correct). But a full re-scan surfaced two more unconfined sinks that need the same treatment before I can register:

**1. Arbitrary file read — `RibbityWildcardResolver`** (`ribbity_wildcards.py:85`): a wildcard token from the `text` STRING is used as `name` in `wildcards_dir / name` → `open(path)` (`:88`). The name regex allows `/` and `\`, and `Path / <absolute>` discards the left operand — so a token like `__/etc/passwd__` resolves to `/etc/passwd` and its contents are returned as node output. No `..`/absolute rejection, no `commonpath`.

**2. Path traversal write — `FrogImagePicker`** (`ribbity_image_picker.py:54`): `_temp_dir(node_id)` joins the unsanitized `node_id` (`unique_id` / `UNIQUE_ID`) onto the temp root as `base / "frog_picker" / str(node_id)` → `mkdir(parents=True)`, then `_save_frames` writes image files there. `node_id` is attacker-influenceable via a crafted graph, so a `../`-laden id escapes the temp root.

Please confine both — reject `..`/absolute and gate the wildcard name and `node_id` with `realpath`+`commonpath` against their intended roots (same pattern you already applied to `output_path`). I'll re-scan once addressed.

RabbitThatIsPink:
Fixes should have been applied now.

ltdrdata:
Thanks — wildcard read and image-picker write are both contained now. Registered to the manager node list.

One recommendation: the Ollama endpoint is hardcoded to a specific LAN address, so this node sends the user's prompt tags there instead of their local Ollama.

- `frog_llm_prompt_refiner.py:19` → `http://192.168.0.194:11434/...`
- `web/frog_llm_control.js:14` → `http://192.168.0.194:11434`

Worth pointing both back to `localhost`, as the docstring and the sibling nodes already do.

PR #2952, opened 2026-06-04
Add Singularity R178 Tail 5 continuation node

DIIIUA:
## Summary
- Add/update the Singularity custom node listing.
- Current listing points to the R178 public release: `Singularity R178 Tail 5 Continuation Gate`.
- R178 adds Tail 5 Continuation Gate for Wan I2V/V2V cascade continuation: pause between cascade segments, inspect source plus five tail-frame candidates, choose the continuation frame, resume the same run, save one stitched video, and write detailed reports.
- Public defaults remain conservative: Observe Only math, Report Only prompt analysis, no prompt text injection, and research modes kept out of the default path.

## Validation
- `python -m json.tool custom-node-list.json`
- `python -m json.tool extension-node-map.json`
- PR branch merged/rebased against current `upstream/main`; diff against upstream only touches the Singularity listing/map entries.
- Upstream conflict resolved.


ltdrdata:
Thanks for the submission. The registered node (`Singularity` / `SingularityCascadeSimple`) looks clean — its writes stay under `folder_paths.get_output_directory()` and it exposes no free-STRING path input. But I have to hold registration over one thing in the shipped code.

**Unconfined file-write class (currently unmapped).** `EventSaveReportToFile` (`nodes.py:1409`) takes a free STRING `output_directory` and writes with no containment:

```python
if output_directory and str(output_directory).strip():
    out_dir = Path(str(output_directory)).expanduser()   # absolute / ~ / ../ all honored
...
out_dir.mkdir(parents=True, exist_ok=True)
path.write_text(str(report or ""), encoding="utf-8")     # attacker-controlled dir + content
```

No `realpath`+`commonpath` confinement, no `..` rejection, and `expanduser()` additionally allows `~` and absolute paths — so an `output_directory` like `/etc/cron.d` or `../../..` creates directories and writes a fully attacker-controlled `.md` anywhere the server can write.

It isn't in `NODE_CLASS_MAPPINGS` today, so it isn't reachable from a workflow as shipped — but it's a complete, functional node one mapping line away from being live, which is why I can't register the pack with it present. Please remove it (or apply the same output-dir confinement the `Singularity` node already uses), then re-request review and I'll re-scan.

DIIIUA:
Fixed in upstream commit [`c143d15`](https://github.com/DIIIUA/ComfyUI-Event-Equality-Core/commit/c143d15).

I removed `EventSaveReportToFile` entirely from the shipped source rather than preserving an unused write surface. It was unmapped and unreferenced, so removal does not change the registered `Singularity` runtime behavior.

Verification performed against the clean public R178 worktree:

- `EventSaveReportToFile` is absent;
- no free `output_directory` STRING widget remains in `nodes.py`;
- `NODE_CLASS_MAPPINGS` still exposes exactly `Singularity`;
- Python compile/AST checks pass.

Please re-scan the updated `master`. Thank you for the precise report.

ltdrdata:
Re-scanned — the fix looks good. `EventSaveReportToFile` is gone from the shipped source, `NODE_CLASS_MAPPINGS` exposes only `Singularity`, and the full write/read sweep found no unconfined sink. Thanks for removing it cleanly.

Registered manually (the branch had a merge conflict against the current list). It's now live in the default channel. Closing this PR — no further action needed.

PR #2815, opened 2026-04-21
Add ComfyUI-GU_Nodepack

alexguryev:
Register [ComfyUI-GU_Nodepack](https://github.com/alexguryev/ComfyUI-GU_Nodepack) in `custom-node-list.json`.

## About the pack

27+ utility nodes for ComfyUI covering:

- LoRA management (stack, randomizer, motion LoRA, trigger-words filter)
- Image annotation (label overlay with configurable font/position)
- Project organization (project paths, naming conventions, file helpers)
- Sampling utilities (indexed sampler/scheduler pickers, resolution presets, sides pack/unpack)
- System integration (timer, save-image-info, get-node-active, switch-any-by-index)

## Metadata

- Repository: https://github.com/alexguryev/ComfyUI-GU_Nodepack
- License: MIT
- Python: >=3.10
- Version: v2.6.0 (tagged)
- pyproject.toml with [tool.comfy] registered on registry.comfy.org (publisher `alexguryev`)

## Verification

- JSON validated with `python -m json.tool` after edit (4469 entries, new entry appended at the end).
- Added the entry at the end of `custom_nodes` array; no existing entries reordered.

PR #2954, opened 2026-06-05
Add ComfyUI-GU_Nodepack

alexguryev:
Register [ComfyUI-GU_Nodepack](https://github.com/alexguryev/ComfyUI-GU_Nodepack) in `custom-node-list.json`.

> Re-submission of #2815, closed for security review. Security rationale is documented
> in the README (added in
> [d91cc62](https://github.com/alexguryev/ComfyUI-GU_Nodepack/commit/d91cc62))
> and summarized below.

## About the pack

27+ utility nodes for ComfyUI covering:

- LoRA management (stack, randomizer, motion LoRA, trigger-words filter)
- Image annotation (label overlay with configurable font/position)
- Project organization (project paths, naming conventions, file helpers)
- Sampling utilities (indexed sampler/scheduler pickers, resolution presets, sides pack/unpack)
- System integration (timer, save-image-info, get-node-active, switch-any-by-index)

## Security rationale

The previous PR was closed because automated scanning flagged `aiohttp.web` import in
`prestartup_script.py`. Here is the full rationale (also in README → Security Notes):

**`prestartup_script.py` imports `aiohttp.web`** — used solely to monkey-patch
`WebSocketResponse.prepare` to hook the moment a browser tab connects to ComfyUI.
The hook refreshes the LoRA file list and resets the seed-file flag.
**No custom HTTP routes or endpoints are registered.**

The patch is applied defensively: checks the method signature before installing,
wraps the original in try/except, falls back gracefully if the aiohttp API changes.

Other patterns confirmed absent:
- No `subprocess` or `exec()` calls
- No `pickle` usage
- JS `fetch()` calls only load the nodepack's own static `.txt` config files — standard ComfyUI extension pattern

## Metadata

- Repository: https://github.com/alexguryev/ComfyUI-GU_Nodepack
- License: MIT
- Python: >=3.10
- Version: v2.6.0 (tagged)
- `pyproject.toml` with `[tool.comfy]`, registered on registry.comfy.org (publisher `alexguryev`)

## Verification

- JSON validated with `python -m json.tool` after edit.
- No existing entries reordered.

### DIFF
--- a/custom-node-list.json
+++ b/custom-node-list.json
@@ -52986,6 +52986,17 @@
             ],
             "install_type": "unzip",
             "description": "This is a node to convert an image into a CMYK Halftone dot image."
+        },
+        {
+            "author": "Alexander Guryev",
+            "title": "ComfyUI-GU_Nodepack",
+            "id": "comfyui-gu-nodepack",
+            "reference": "https://github.com/alexguryev/ComfyUI-GU_Nodepack",
+            "files": [
+                "https://github.com/alexguryev/ComfyUI-GU_Nodepack"
+            ],
+            "install_type": "git-clone",
+            "description": "Several utility nodes for ComfyUI: LoRA management, image annotation, project organization, sampling utilities, system integration."
         }
     ]
 }


alexguryev:
Updated to v2.6.1: removed `aiohttp.web` import from `prestartup_script.py`. LoRA list refresh now handled via `INPUT_TYPES()` (standard ComfyUI pattern, called on each `/object_info` request). No WebSocket patching remains in the codebase.

ltdrdata:
Thanks for the resubmission and the README notes. The prestartup WebSocket monkey-patch rationale is fine, and I confirmed there's no subprocess/exec/pickle. But I have to hold this — two nodes write/read attacker-controlled paths with no containment.

**1. Arbitrary file write — `G_SaveImgInfo`** (`nodes.py:194`):

```python
if len(file_name) > 0:
    fpath = make_unique_filename(os.path.join(folder_paths.get_output_directory(), file_name + ".txt"))
    file = open(fpath, "w")
```

`file_name` is joined straight onto the output dir with no `basename`, no `realpath`+`commonpath`, no `..` rejection. `make_unique_filename` only de-duplicates — it doesn't confine. `file_name = "../../../../tmp/evil"` writes a `.txt` outside the output directory.

**2. Path traversal (write + read) — `G_SeedList`** (`nodes.py:352`):

```python
if len(project) > 0:
    tmp = os.path.join(tmp, "(work)", project)   # project not sanitized
    os.makedirs(tmp, exist_ok=True)
fname = os.path.normpath(os.path.join(tmp, f"{fn}.txt"))
... open(fname, "r") ... open(fname, "w")
```

`filename` goes through `safe_string`, but `project` does not. `os.path.normpath` doesn't confine — `project = "../../../../tmp/x"` escapes the output dir, so `os.makedirs` + `open(fname, "w")` write, and `open(fname, "r")` reads, at an attacker-chosen location.

Both are remotely reachable: `POST /prompt` accepts arbitrary graphs cross-origin with no auth/CSRF token, and the write is server-side so CORS doesn't mitigate it.

Please confine both `file_name` and `project` to the output root (`realpath`+`commonpath`, reject `..`/absolute — the same `safe_string` treatment you already apply to `filename`), then re-request review and I'll re-scan.

alexguryev:
Fixed in v2.6.6.

Both paths are now confined to the output root via a shared helper, `SafeOutputPath` in `functions.py`:

- the untrusted relative path is split into segments; any segment that is `..`, or contains `:` (drive prefix / NTFS ADS), or collapses to nothing after trailing-dot/space stripping, **rejects the whole path** (returns `None`) — it is never silently rewritten into something else
- leading `/` (absolute path) is rejected
- characters illegal on Windows and control chars are replaced within a segment, and Windows reserved device names are prefixed
- finally the joined path goes through `os.path.realpath` + `os.path.commonpath` against the resolved output root (symlinks included); anything landing outside is rejected

`G_SaveImgInfo` (`nodes.py:198`) passes `file_name` through it, and `G_SeedList` (`nodes.py:364`) passes the combined `(work)/<project>/<filename>` through it, so `project` is now covered too. On rejection both nodes log an error and skip the file operation entirely instead of falling back to a default path.

Directory creation goes through `SafeMakeDirs`, which re-verifies containment *after* `makedirs` to close the TOCTOU window where a symlink could be swapped in between the check and the open.

Re-requesting review — thanks for the detailed writeup, the `realpath`+`commonpath` pointer was exactly what was missing.

ltdrdata:
Re-scanned — the fix looks good. Both `G_SaveImgInfo` and `G_SeedList` now route through `SafeOutputPath` (segment validation + realpath/commonpath, rejecting `..`/absolute/drive prefixes), and the full sweep found no other unconfined sink. Thanks for the thorough hardening.

Registered manually (the branch had a merge conflict against the current list). It's now live in the default channel. Closing this PR — no further action needed.

PR #2966, opened 2026-06-07
Update Realtime LoRA Toolkit description

shootthesound:
Updates the description for the existing `comfyui-realtime-lora` entry to reflect the current node set. The Clippy Reloaded, Image of the Day, and Model Diff to LoRA utility nodes are now published as standalone nodes, so the description now focuses on the training / analysis / editing toolkit.


ltdrdata:
Thanks — I've applied the updated `description` from this PR to the existing `Realtime LoRA Toolkit` entry in `custom-node-list.json` directly.

Closing this PR — the entry is now updated.

PR #2972, opened 2026-06-10
fix(git_compat): harden pygit2 fallback path; prefer system git when available

Kosinkadink:
## Summary

Hardens the **pygit2 fallback path** in `git_compat.py`. Paired with Comfy-Org/Comfy-Desktop#1038, which is the actual fix for the Desktop update/migration failure (Comfy-Desktop#1036).

> Targets `manager-v4` (not shipped on `main` yet).

### Context
The bundled pygit2/libgit2 honors the user's global git config but ships **HTTPS-only (no SSH transport)**. A global `insteadOf` rewrite (`https://github.com/` ? `git@github.com:`) makes pygit2 fetch/clone fail with `authentication required but no callback set`.

Manager's backend selection already does the right thing and is **unchanged** here:

```
USE_PYGIT2 = (CM_USE_PYGIT2 == '1')   # only when explicitly forced
if not USE_PYGIT2:
    try: git --version                 # probe system git via GitPython
    except: USE_PYGIT2 = True          # pygit2 only when system git is absent
```

Since Comfy-Desktop#1038 stops forcing `CM_USE_PYGIT2`, Manager now **prefers system git (GitPython) whenever it's present** - which honors the user's full config (`insteadOf`, proxy, ssh keys, credential helpers) and sidesteps the bug entirely. There's no real reason to prefer pygit2 when system git exists, so we don't.

### What this PR changes (fallback path only)
pygit2 is now reached only as a no-system-git fallback (or when a developer sets `COMFY_FORCE_PYGIT2=1`). To keep that path robust:

- **Ignore system/global/XDG git config** for libgit2 ops (in-memory `search_path` blanking). Protects the no-git box that still has a stray `~/.gitconfig` with an `insteadOf` rewrite (e.g. from GitHub Desktop or synced dotfiles), which would otherwise reproduce the identical failure.
- **Preserve corporate `http.proxy`** - snapshot it before blanking and pass it explicitly to clone/fetch, so config blanking doesn't strip a corporate proxy.

All changes live inside the `if USE_PYGIT2:` path, so they are inert whenever system git is available.

### Not destructive
Removed the earlier on-disk remote-URL rewrite (`_normalize_remote_urls`): persistently rewriting a repo's stored `origin` is risky if interrupted and unnecessary given the config blanking. Everything here is in-memory only.

### DIFF
--- a/comfyui_manager/common/git_compat.py
+++ b/comfyui_manager/common/git_compat.py
@@ -447,6 +447,28 @@ class _Pygit2Repo(GitRepo):
         self._repo = _pygit2.Repository(git_dir)
         self._working_dir = repo_path
 
+    def _fetch_remote(self, remote, refspecs=None):
+        """Fetch *remote* over the preserved proxy, transparently rewriting an
+        SSH-form origin to anonymous HTTPS in memory.
+
+        The bundled pygit2 has no SSH transport, so a stored `git@host:...`
+        origin would fail with an auth error. When the URL is SSH-form we fetch
+        through an in-memory anonymous remote over HTTPS, leaving `.git/config`
+        untouched (no on-disk rewrite).
+        """
+        https_url = _to_https_url(remote.url)
+        if https_url != remote.url:
+            anon = self._repo.remotes.create_anonymous(https_url)
+            anon.fetch(
+                list(refspecs) if refspecs is not None
+                else list(remote.fetch_refspecs),
+                proxy=_HTTP_PROXY,
+            )
+        elif refspecs is not None:
+            remote.fetch(list(refspecs), proxy=_HTTP_PROXY)
+        else:
+            remote.fetch(proxy=_HTTP_PROXY)
+
     @property
     def working_dir(self):
         return self._working_dir
@@ -503,7 +525,7 @@ class _Pygit2Repo(GitRepo):
         remote = self._repo.remotes[name]
 
         def _pull():
-            remote.fetch(proxy=_HTTP_PROXY)
+            self._fetch_remote(remote)
             branch_name = self.active_branch_name
             branch = self._repo.branches.get(branch_name)
             if branch and branch.upstream:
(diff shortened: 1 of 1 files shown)


Kosinkadink:
Follow-up commit: **preserve corporate `http.proxy`**.

The bundled pygit2 ships with HTTPS but **no SSH transport** (`pygit2.features & Feature.SSH == False`), so a user's `insteadOf` https->ssh rewrite can never succeed regardless of credentials - forcing anonymous HTTPS is the only viable path. The one corporate setting we'd otherwise lose by blanking the config is the proxy, so we now snapshot `http.proxy` from the global config **before** blanking the search path and pass it explicitly (`proxy=`) to every fetch/clone/list_heads call.

Verified end-to-end against pygit2 1.19.2: proxy is read correctly, `insteadOf` is still neutralized, and fetch/clone with the proxy value succeed.

Known limitation: `http.sslVerify=false` (corporate MITM with an untrusted CA) is still dropped. That's a deliberate choice - auto-disabling TLS verification is security-sensitive. Can add it (snapshot + `certificate_check` callback) if needed.

Kosinkadink:
## Updated approach: prefer system git; no on-disk remote rewrites

This PR is paired with Comfy-Desktop#1038. The overall strategy changed away from rewriting repo state and toward preferring **system git** (which honors the user's full git config) while keeping the bundled pygit2 as a fallback.

### Root cause (Comfy-Desktop issue #1036)
The bundled pygit2/libgit2 honors the user's global git config but ships **HTTPS-only (no SSH transport)**. A global `insteadOf` rewrite (`https://github.com/` ? `git@github.com:`) makes pygit2 fetch/clone fail with `authentication required but no callback set`. Because Launcher was forcing `CM_USE_PYGIT2=1` for standalone installs, Manager v4 hit the same failure on update/clone.

### What changed here
- **Removed `_normalize_remote_urls()`**: no longer persistently rewrites a repo's SSH `origin` to HTTPS on disk. Mutating on-disk repo state is risky if interrupted (power loss/crash), and it's unnecessary - the pygit2 backend already blanks libgit2's config search path so anonymous-HTTPS fetch works without touching the stored remote.
- Backend selection (unchanged, but now actually relied upon): Manager **prefers GitPython/system git** when a system git is present (honoring `insteadOf`, proxies, ssh keys, credential helpers) and only uses the bundled pygit2 when system git is absent or `CM_USE_PYGIT2=1` is set.
- On the Launcher side, `CM_USE_PYGIT2=1` is now only injected when a developer sets `COMFY_FORCE_PYGIT2=1`, so Manager defaults to system git when available.

The libgit2 config-search-path blanking and `http.proxy` preservation remain for the pygit2-only path (system git absent).

Note: targets the `manager-v4` line (not shipped on main yet).

PR #2992, opened 2026-06-15
Add ComfyUI-SCAIL-Pose2 custom node

rookiestar28:
## Summary

Adds ComfyUI-SCAIL-Pose2 to `custom-node-list.json` as a git-clone installable custom node.

The entry follows the existing rookiestar28 records and points to:

- https://github.com/rookiestar28/ComfyUI-SCAIL-Pose2

## Scope

ComfyUI-SCAIL-Pose2 provides SCAIL and SCAIL-2 pose/mask preprocessing nodes for ComfyUI, including DWPose-compatible pose data, NLF pose renders, condition metadata, and WanVideoWrapper adapter payloads.

## Validation

- `python json-checker.py custom-node-list.json`
- `git diff --check -- custom-node-list.json`

### DIFF
--- a/custom-node-list.json
+++ b/custom-node-list.json
@@ -37654,6 +37654,17 @@
             "description": "Your own personal AIGC Factory. Any picture. Any reel. The Comfy way. ©️"
         },
         {
+            "author": "rookiestar28",
+            "title": "ComfyUI-SCAIL-Pose2",
+            "id": "comfyui-scail-pose2",
+            "reference": "https://github.com/rookiestar28/ComfyUI-SCAIL-Pose2",
+            "files": [
+                "https://github.com/rookiestar28/ComfyUI-SCAIL-Pose2"
+            ],
+            "install_type": "git-clone",
+            "description": "SCAIL and SCAIL-2 pose/mask preprocessing for ComfyUI, including DWPose-compatible pose data, NLF pose renders, condition metadata, and WanVideoWrapper adapter payloads."
+        },
+        {
             "author": "MoonMoon82",
             "title": "ClipVision_Tools",
             "reference": "https://github.com/MoonMoon82/ClipVision_Tools",


ltdrdata:
Thanks for the submission! A pre-registration security review flagged a path-handling issue that should be resolved before this can be added to the default channel.

In **`SaveNLFPosesAs3D`**, the `filename_prefix` input (a free-text widget) is joined onto the output directory without containment, so a crafted value can make the `.glb` write land **outside** ComfyUI's `output/` folder. ComfyUI's built-in save nodes avoid this by routing the prefix through `folder_paths.get_save_image_path()`, which strips traversal and confines the write to the output tree.

**Suggested fix:** replace the raw `os.path.join(get_output_directory(), filename_prefix)` with `folder_paths.get_save_image_path(filename_prefix, folder_paths.get_output_directory())` (or at minimum `os.path.basename()` plus a `commonpath` containment check).

Once updated, reply here and we'll re-review. Holding the PR open in the meantime — thanks for helping keep the ecosystem safe!

rookiestar28:
> Thanks for the submission! A pre-registration security review flagged a path-handling issue that should be resolved before this can be added to the default channel.
> 
> In **`SaveNLFPosesAs3D`**, the `filename_prefix` input (a free-text widget) is joined onto the output directory without containment, so a crafted value can make the `.glb` write land **outside** ComfyUI's `output/` folder. ComfyUI's built-in save nodes avoid this by routing the prefix through `folder_paths.get_save_image_path()`, which strips traversal and confines the write to the output tree.
> 
> **Suggested fix:** replace the raw `os.path.join(get_output_directory(), filename_prefix)` with `folder_paths.get_save_image_path(filename_prefix, folder_paths.get_output_directory())` (or at minimum `os.path.basename()` plus a `commonpath` containment check).
> 
> Once updated, reply here and we'll re-review. Holding the PR open in the meantime — thanks for helping keep the ecosystem safe!

Thanks for the review! This has been fixed. `SaveNLFPosesAs3D` now routes `filename_prefix` through `folder_paths.get_save_image_path()` and constructs the GLB filename from the sanitized basename it returns. Traversal and absolute-path prefixes are rejected before the exporter is invoked, while legitimate nested output prefixes remain supported. Regression tests and the full repository validation suite pass.

PR #2984, opened 2026-06-12
Add comfyui-ZoeyTool to custom node list

liangzoey:
Add @liangzoey/comfyui-ZoeyTool custom node to ComfyUI Manager.

- GitHub: https://github.com/liangzoey/comfyui-ZoeyTool
- 19 custom nodes for image editing, lighting control, video processing, prompt generation/translation, and more.
- Properly structured with NODE_CLASS_MAPPINGS, WEB_DIRECTORY, and __init__.py exports.

PR #3002, opened 2026-06-16
Update comfyui-ZoeyTool entry: add pip dependencies

liangzoey:
Add pip dependencies (`opencv-python-headless`, `transformers`) to the existing comfyui-ZoeyTool entry so they are automatically installed by ComfyUI Manager.

Also update the description to reflect the latest mask bounding box node features (lighting prompt output).


ltdrdata:
Hi @liangzoey — this duplicates #2984 (same `comfyui-ZoeyTool` entry) and is built on an older `custom-node-list.json` snapshot with a whole-file reindent, so it can't be merged as-is (it would revert unrelated entries). The pip deps here are also already in your repo's `requirements.txt`, so Manager installs them automatically — an explicit `pip` field isn't needed.

Let's keep everything on #2984 (I've left the security review there). Closing this one as a duplicate — thanks!

PR #3013, opened 2026-06-21
Add SpriteSheetExtractor node

ugotworms:
## Add SpriteSheetExtractor custom node

Adds two nodes for extracting sprite sheets from WAN (or any batch) video output:

- **SpriteSheetExtractor** — Samples evenly-spaced frames, removes background via per-frame YCbCr chroma keying with flood-fill, applies palette lock for colour drift correction, and saves a horizontal RGBA sprite sheet PNG. Includes an interactive "Tweak Tolerance" panel for live adjustment without re-running generation.
- **SpriteSheetPreview** — Displays the sprite sheet as a looping pixel-art animation with checkerboard background for transparent areas.

**Details:**
- `install_type`: `git-clone`
- No extra pip dependencies (uses PyTorch, numpy, Pillow — all ComfyUI built-ins)
- NODE_CLASS_MAPPINGS: `SpriteSheetExtractor`, `SpriteSheetPreview`
- Category: `image/animation`

**Repo:** https://github.com/ugotworms/comfyui_sprite_sheet_extractor

### DIFF
--- a/custom-node-list.json
+++ b/custom-node-list.json
@@ -53410,6 +53410,20 @@
             ],
             "install_type": "unzip",
             "description": "This is a node to convert an image into a CMYK Halftone dot image."
+        },
+         {
+            "author": "ugotworms",
+            "title": "Sprite Sheet Extractor Nodes for ComfyUI",
+            "id": "sprite-sheet-extractor",
+            "reference": "https://github.com/ugotworms/comfyui_sprite_sheet_extractor",
+            "files": [
+                "https://github.com/ugotworms/comfyui_sprite_sheet_extractor"
+            ],
+            "install_type": "git-clone",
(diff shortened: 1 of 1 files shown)


ltdrdata:
Thanks for the submission. I have to hold this — the interactive save route writes to an attacker-controlled path with no containment.

**`POST /sprite_sheet/save`** (`__init__.py:55`) builds the output filename from a caller-supplied `filename_prefix` (request body) and joins it onto the output directory with no `basename`/`commonpath`/`..` rejection, so a `../`-laden prefix escapes the output directory (`Image.save`). The route is unauthenticated and CSRF-reachable; the write is server-side so CORS doesn't mitigate it.

Please confine the save path to the output directory and reject `..`/absolute. I'll re-scan once addressed.

PR #2987, opened 2026-06-13
Update Deno custom node map

Deno2026:
## Summary

Updates the Deno Custom Nodes entry in `extension-node-map.json` from the old single-node mapping to the current public node list.

The package now exposes these nodes through its repository `node_list.json`, and Comfy Registry version `0.7.31` is active. Without this map update, ComfyUI-Manager still shows the pack as `1 node` even though the current package contains the full public node set.

## Changes

- Update root `extension-node-map.json` for `https://github.com/Deno2026/comfyui-deno-custom-nodes`
- Update `node_db/new/extension-node-map.json` with the same node list
- Add 18 public Deno node class IDs, including Local LLM Loader / Reviewer and Prompt Text

## Validation

- Ran ComfyUI-Manager scanner in scan-only mode against `https://github.com/Deno2026/comfyui-deno-custom-nodes`
- Scanner output found 18 Deno nodes
- `python -m json.tool extension-node-map.json`
- `python -m json.tool node_db/new/extension-node-map.json`
- `python json-checker.py extension-node-map.json`
- `python json-checker.py node_db\new\extension-node-map.json`
- Verified both Deno map entries contain exactly the expected 18 nodes

PR #3021, opened 2026-06-26
Update Deno Custom Nodes LTX AV sampler mapping

Deno2026:
## Summary

- Replace the retired `DenoLTXStepFusedTiledSampler` mapping with `DenoLTXAVStepFusedTiledSampler` for Deno Custom Nodes.
- Keep `DenoLTXTiledSpatialUpscaler` unchanged.

## Context

Deno Custom Nodes v0.7.64 keeps the LTX tiled spatial upscaler and publishes the AV-aware tiled sampler as the public sampler node. The old video-only sampler is no longer registered publicly.

### DIFF
--- a/extension-node-map.json
+++ b/extension-node-map.json
@@ -7474,7 +7474,7 @@
             "DenoLTXMultiLoraLoader",
             "DenoLTXPromptGuide",
             "DenoLTXSequencer",
-            "DenoLTXStepFusedTiledSampler",
+            "DenoLTXAVStepFusedTiledSampler",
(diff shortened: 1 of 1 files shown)


ltdrdata:
Hi @Deno2026 — same as #2987: `extension-node-map.json` isn't hand-edited — it's extracted and registered automatically by a scanner script that reads the node classes directly from each repo. So manual edits to it aren't merged; your pack's node classes are reflected automatically on the next scan. No action is needed on your end.

Closing this for that reason — thanks!

PR #3025, opened 2026-06-28
Add xhhhhh Toolkit custom nodes

xxxhh123fd:
添加 xhhhhh 工具集，包含7个实用ComfyUI自定义节点：文本一键粘贴、连线透明控制、选择任意、图像裁剪加载器、出图尺寸比例、提示词行、保存图像

### DIFF
--- a/custom-node-list.json
+++ b/custom-node-list.json
@@ -1,6 +1,16 @@
 {
     "custom_nodes": [
         {
+            "author": "xhhhhh",
+            "title": "xhhhhh 工具箱",
+            "reference": "https://github.com/xxxhh123fd/ComfyUI-xhhhhh-Toolkit",
+            "files": [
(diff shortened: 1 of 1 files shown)


ltdrdata:
Thanks for the submission. I have to hold this — the save node writes to an attacker-controlled path with no containment. (Node-widget strings are still remotely reachable via `POST /prompt` CSRF; the write is server-side so CORS doesn't mitigate it.)

**`SaveImage_XH`** (`save_image.py:84`) — a free STRING prefix yields a sub-path that is joined onto the output dir with only `.strip()`/`.rstrip('/')` (no `..` rejection, no `commonpath`), so a `../`-laden value escapes the output tree at `os.makedirs` + `save`.

Please confine the path to `folder_paths.get_output_directory()` and reject `..`/absolute. I'll re-scan once addressed.

(Separately: the PR diff rewrites the entire `custom-node-list.json` — please rebase so the diff is just your one new entry.)

PR #3029, opened 2026-06-29
Register ComfyUI Dynamic Text-to-Combo Generator custom node

gh055:


ltdrdata:
Thanks — I've applied the updated `title` and `description` from this PR to the existing `ComfyUI Dynamic Text-to-Combo Generator` entry in `custom-node-list.json` directly.

Closing this PR — the nodepack is now registered with the updated metadata.

PR #3043, opened 2026-07-02
Add Krea2 BBOX Prompter Suite

ukr8b3g-cmyk:
Register Krea2 BBOX Prompter Suite in ComfyUI-Manager.

Repository: https://github.com/ukr8b3g-cmyk/Krea2-BBOX-Prompter

This custom node suite provides a Krea2-focused BBOX canvas, prompt builder, export node, and prompt effect presets for ComfyUI.


ltdrdata:
Thanks for the contribution! Before this can be merged, the following Non-English UI strings need to be addressed:

1. **i18n LOCALES table** — `web/krea2_element_framing_v1.js` ships a full `ja:` locale block (lines ~2931-2987) with native Japanese UI labels/tooltips that auto-bind by `navigator.language`. Please make it opt-in per-user rather than auto-binding.

2. **Language-select combos with native names** — `nodes_element_framing.py:364` and JS dropdowns (lines ~4903, ~5803) expose `"Japanese 日本語"`. Please use ASCII-only language codes (`"ja"`) or English names (`"Japanese"`) — the native-language name in the value is blocked by policy for consistency.

ukr8b3g-cmyk:
Implemented all requested review changes:

- Removed browser/ComfyUI locale auto-detection; Japanese UI is now enabled only when the user explicitly selects “Japanese”.
- Replaced the non-ASCII “Japanese 日本語” language values with the ASCII/English value “Japanese”.
- Updated the PR branch to the latest ComfyUI-Manager main and resolved the custom-node-list.json conflict while preserving both entries.
- Replaced the existing “YOUR_COMFY_PUBLISHER_ID” placeholder in the Undulating Glitch entry with “jsterlingvids”, matching the owner in its repository URL.

The latest fix is included in commit 1f288b2. Please re-review.

PR #3062, opened 2026-07-07
Added two node packs to custom-node-list.json

CovertBannana:
ComfyUI_CivitAI_ImagePlus and Unified node pack ComfyUI-ImagePlus.


ltdrdata:
I've registered `ComfyUI-ImagePlus` in `custom-node-list.json`.

`ComfyUI_CivitAI_ImagePlus` is held back due to a server-side request forgery (SSRF) issue:

- `civitai_gallery_backend.py:400` — `_maybe_fetch_tensor_from_image_url` fetches `image_url` via `urllib.request.urlopen` in a fallback branch with no host validation. The allowlisted branch (`_is_allowed_image_url`) only runs when an API key is present and the URL passes the allowlist; otherwise control falls through to the unguarded fallback, so a disallowed URL is still fetched. The URL is attacker-influenceable via the unauthenticated `POST /civitai_gallery/set_preview` route (no host validation on `body['url']`), letting a request to the ComfyUI API make the server fetch arbitrary/internal hosts (e.g. `http://169.254.169.254/`).

Please apply the `_is_allowed_image_url()` check to the fallback branch as well (and validate the host on the `set_preview` route before storing the URL). I'll re-evaluate `ComfyUI_CivitAI_ImagePlus` once the fallback enforces the allowlist.

CovertBannana:
Thanks for catching the SSRF issue.

I’ve pushed fixes to ComfyUI_CivitAI_ImagePlus.

Changes made:
- `_maybe_fetch_tensor_from_image_url()` now validates image URLs with `_is_allowed_image_url()` before every server-side fetch path, including the fallback branch.
- `/civitai_gallery/set_preview` now validates `body["url"]` before storing preview state.
- `page_url` is also validated with `_is_allowed_page_url()`.
- Server-side image fetches now use parsed hostname checks instead of raw `netloc` matching.
- Redirects are followed only when each redirect target remains on the CivitAI allowlist.

Please re-evaluate when you have a chance.
Thanks.

CB.

ltdrdata:
Both nodepacks from this PR are now registered — `ComfyUI-ImagePlus` and `ComfyUI_CivitAI_ImagePlus`. I've updated the existing `ComfyUI_CivitAI_ImagePlus` entry with the title/description and author from this PR, and re-verified the SSRF fixes at current head (allowlist enforced on every fetch path, redirects re-validated).

Closing this PR — the nodepacks are registered. Thanks!

PR #3063, opened 2026-07-07
Add ComfyUI-NukeLink

drberkowitz:
## Summary
Adds ComfyUI-NukeLink to the ComfyUI Manager custom node list.  ComfyUI-NukeLink bridges Foundry's Nuke and ComfyUI on the same machine.

## Extension Information
- **Name:** ComfyUI NukeLink
- **Author:** drberkowitz
- **Repository:** https://github.com/drberkowitz/ComfyUI-NukeLink
- **Install type:** git-clone
- **Nodes:** NukeLink - Read, NukeLink - Write, NukeLink - Path Builder

## Features
* Send one or more Nuke Read nodes to ComfyUI with knob values intact: file path, colorspace, frame range, and missing frames mode
* Automatic Path Builder node placement, pre-populated with output location, and shot name
* Shot name and output location derived from pipeline environment variables or script filename
* Send ComfyUI outputs back to Nuke as Read nodes via right-click menu on any NukeLink - Write node
* Sequence and still image preview with right-click menu playback controls on Read and Write nodes
* Missing frames handling: black, hold, nearest, and error modes
* Colorspace support: raw, sRGB, linear, and ACEScg
* Multiple simultaneous Nuke instances
* Standard, portable, and desktop ComfyUI installs supported

### DIFF
--- a/custom-node-list.json
+++ b/custom-node-list.json
@@ -54176,6 +54176,17 @@
             ],
             "install_type": "git-clone",
             "description": "Native ComfyUI nodes for EverAnimate-style chunked WanAnimate workflows, including master/chunk settings, chunk calculation, image carry-over, and boundary color correction."
+        },
+        {
+            "author": "drberkowitz",
+            "title": "ComfyUI-NukeLink",
+            "id": "comfyui-nukelink",
+            "reference": "https://github.com/drberkowitz/ComfyUI-NukeLink",
+            "files": [
+                "https://github.com/drberkowitz/ComfyUI-NukeLink"
+            ],
+            "install_type": "git-clone",
+            "description": "A bridge between Nuke and ComfyUI on the same machine. Send Read nodes from Nuke directly to the ComfyUI canvas with knob values intact, and return rendered outputs back to Nuke as Read nodes."
         }
     ]
 }


drberkowitz:
Hi @ltdrdata, first time contributor here, just checking in on this PR.  Wanted to make sure everything looks right on my end. Seems to have passed CodeRabbit with no issues flagged. More than happy to make any changes if something's needed. Thank you!

ltdrdata:
Hi @drberkowitz — thanks for the PR! Holding this for a couple of security fixes first:

1. **`NukeLinkWrite`**: `file_path` isn't confined to the output dir — absolute/`..` paths write anywhere. Please add `realpath`+`commonpath` containment.
2. **`/nukelink/openinfolder`**: this unauthenticated route passes the caller's `path` straight to `open`/`explorer`. On macOS `open` will **launch a `.app`** — an app-launch capability a "reveal in folder" endpoint shouldn't have. Best fix: only ever pass a **folder** to `open` (and reject `.app` bundles, which are technically directories), so the endpoint can only open a directory and never launch anything. Same containment applies to the other path-taking routes.

(Note: the CodeRabbit review above only covers this PR's `custom-node-list.json` diff, not your node repo — the above is from a separate review of `ComfyUI-NukeLink`.)

Happy to re-review once these are in. Thanks!

drberkowitz:
Hi @ltdrdata , sorry about the delay! I ended up expanding the tool a little bit while I was working on your notes. Both issues should be fixed:

- Write node paths now use strict containment checks on relative paths, ensuring that any relative inputs using dot-dot sequences to escape the output directory are rejected while preserving legitimate absolute paths.
- The open-in-folder route now only ever passes a folder to open/explorer, and .app bundles are explicitly rejected so they can't be opened through it. The same path containment check now applies to the other path-taking routes as well.

Let me know if this covers what you were looking for, thank you!

ltdrdata:
Thanks for the updates — the relative-path check is correct now, but the write still isn't safe. In `utils.py` `resolve_file_path` (17-29) the `realpath`+`commonpath` check only runs for relative paths; absolute paths (and `~/…`/`$HOME/…`, which expand to absolute) pass through unconfined, so `NukeLinkWrite` can still write anywhere. These are open routes with no access control, so any remote caller can drive them. Please apply the same `realpath`+`commonpath==base` check to absolute paths too.

Same containment/gating still needed on:
- `/openinfolder` — launches a file manager on any absolute dir
- `/viewimage`, `/viewvideo` — no containment; return any file's contents to any caller
- `/fileexists` — leaks existence of arbitrary paths
- `/sendtonuke` — validate `nuke_port` to the 54321-54330 range like `/pingport`

Happy to re-review once each path input is confined.

drberkowitz:
Thanks for the detailed follow-up @ltdrdata , just got another update out!

**Fixed:**

- `/sendtonuke` now validates `nuke_port` against the same `54321-54330` range `/pingport` already enforces.
- `~`/`$HOME`-style paths are resolved to absolute before the relative/absolute branch decision, so they can no longer bypass the relative-path containment check.
- `resolve_file_path` now rejects any path containing a `..` component. This closes traversal-by-string and applies uniformly to relative and absolute input.
- `/fileexists` had no caller anywhere in the codebase, front end or back end, so I got rid of it instead of patching it.
- `/viewimage` and `/viewvideo` now route their `filename` param through `resolve_file_path` before touching disk, so they inherit the same traversal/symlink protection as the write path.
- `/openinfolder` already resolves through the same function and rejects `.app` bundles; the subprocess calls are list-form with no `shell=True`, so they're not injectable via the path value. Don't think this one needs any more changes, but happy to hear otherwise.

**Open question:**

None of the above restricts which absolute destination a Read or Write can target, they only close off ways of disguising or redirecting a path string. That's by design, and I want to be straight about it instead of implying it's more settled than not.

NukeLink's entire purpose is reading from and writing to wherever a VFX artist or studio's pipeline is built, something like `E:/Shows/BigShow/Shots/BS-101-010/`, which by definition isn't knowable in advanced and doesn't have a fixed root to confine against. In the VFX space, tools reaching into absolute, uncontained paths is common, almost a prerequisite for anything meant to connect to an actual pipeline.

If the concern is specifically that NukeLink surfaces this over HTTP routes rather than keeping it entirely inside node execution, that's a whole other situation I'd genuinely like to understand better, and if there's a way to address the network-exposure side of this without restricting the destination path itself, I'd be very open to it. I want to get this right, but as best I understand the ask, full path containment would knee-cap the tool's core function rather than secure it. Let me know what I'm missing and if there's anything I'm misunderstanding!

ltdrdata:
Thanks for the quick turnaround, and for the detailed write-up of what changed — that made this easy to re-check. A few things clearly landed: `/fileexists` is gone, `openinfolder` now uses a list-form call with no shell so there's no command-building issue there any more, `sendtonuke` validates the port, and `resolve_file_path` now rejects paths containing `..`. Good improvements.

Your open question is the right one to ask, and I think we actually agree — let me answer it directly. You're correct that I am NOT asking you to constrain the pipeline paths. NukeLink reading and writing wherever a studio's pipeline lives (E:/Shows/BigShow/...) is the point of the tool, and the node inputs should stay exactly as they are. Full path containment on the nodes would knee-cap it, and that's not the fix.

The concern is narrower: those arbitrary read / write / open actions are reachable over HTTP routes that have no access control, so any web page the user happens to have open can drive them from the user's own machine. That's the network-exposure side you mentioned, and that's exactly where the fix belongs. A same-machine / loopback IP check does NOT cover this — the request already comes from the user's own machine, so it looks local either way.

So the remedy for each route is to make it local-only and explicitly user-initiated rather than something a background web request can trigger — or to drop the route and keep the behaviour inside node execution, where the node inputs already give you the arbitrary paths you need. Here's what the re-check still sees:

1. `resolve_file_path` (utils.py:26) — the `..` rejection landed, but it still returns any absolute path unchanged, so it gives no confinement for callers that pass an absolute path. Every sink below leans on it.
2. `/nukelink/viewimage` (server.py:600) — an open route with no access control; it reads a caller-supplied absolute path and returns the decoded image, so it can hand back any readable image file on the host. Make this route local-only / user-initiated, or move the read into node execution.
3. `/nukelink/viewvideo` (server.py:321) — same open-route pattern; it transcodes and streams frames from a caller-supplied absolute sequence path with no containment. Same fix.
4. `/nukelink/openinfolder` (server.py:263) — the shell side is fixed, but it still launches the file explorer on a caller-supplied path with no base-dir containment over an open route. Same fix — gate it to a local, user-initiated action.
5. `/nukelink/sendtonuke` (server.py:117) — resolves a caller-supplied absolute path and relays it over the localhost socket with no containment; still reachable over an open route. Same fix.
6. NukeLinkWrite (nukelink_nodes.py:308) — this one writes image bytes and makedirs at a widget-supplied `file_path`. Keeping arbitrary destinations here is fine by design, as long as it's driven by node execution and not reachable from an open route.

To be clear on where the split is: I'm not asking you to restrict which absolute path a Read or Write can target — keep that. I'm asking that these actions not be triggerable by an arbitrary web request. Routing filenames through `resolve_file_path` helps, but as long as it lets absolute paths through and the routes are network-reachable, the exposure is still open.

I'd be glad to look at whichever approach you prefer (local-only routes vs. folding the behaviour back into node execution) — happy to talk it through.

PR #3095, opened 2026-07-20
Add ComfyUI_QBB_LoadHighlight to custom-node-list

rhinoflavored:
## Summary
- Add `ComfyUI_QBB_LoadHighlight` so it can be found and installed via ComfyUI-Manager's default channel list.

## Node
- Repo: https://github.com/rhinoflavored/ComfyUI_QBB_LoadHighlight
- Registry: already published as `qbb-loadhighlight` / DisplayName `QBB LoadHighlight`
- Install type: `git-clone`

## Test plan
- [ ] Search Manager for `QBB LoadHighlight` or `ComfyUI_QBB_LoadHighlight`
- [ ] Install via Manager and confirm plugin loads

### DIFF
--- a/custom-node-list.json
+++ b/custom-node-list.json
@@ -60823,6 +60823,17 @@
             ],
             "install_type": "git-clone",
             "description": "A ComfyUI custom node that randomly selects one input from multiple inputs of the same type."
+        },
+        {
+            "author": "rhinoflavored",
+            "title": "QBB LoadHighlight",
+            "id": "qbb-loadhighlight",
+            "reference": "https://github.com/rhinoflavored/ComfyUI_QBB_LoadHighlight",
+            "files": [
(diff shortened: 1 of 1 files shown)


ltdrdata:
Thanks for the submission. I have to hold this — two export routes write to an attacker-controlled directory with no containment (`qbb_loadhighlight/routes.py`).

**`POST /qbb_loadhighlight/export_to_path`** (`:148`) and **`/export_environment_to_path`** (`:179`) take a free `directory` from the request body and pass it (via `abspath`+`expanduser`, which don't confine) to `os.makedirs` + `open("w")` → the file is written to any absolute path (or `~/…`) the server can reach. The routes are unauthenticated and CSRF-reachable; the write is server-side so CORS doesn't mitigate it.

Please confine `directory` to an intended output root and reject `..`/absolute/`~`. I'll re-scan once addressed.

PR #3101, opened 2026-07-21
Add ComfyUI-prompt-helper

chenr5934-tech:


ltdrdata:
Thanks for the submission. I have to hold this — it bundles the full `sd-webui-prompt-all-in-one` extension and auto-starts a FastAPI server (port 17861) that exposes package installation over the network.

**`POST /physton_prompt/install_package`** (`scripts/on_app_started.py:88` → `packages.py:39`) drives `pip install` from the request body. Allowing package installation to be triggered remotely is prohibited: `pip install <anything>` runs the package's build/`setup.py` at install time, so this is arbitrary code execution on the host by design (an allow-list of names does not change that). Please remove the install route — installation must never be reachable from the network; it can only be local and explicitly user-initiated.

**Path traversal / arbitrary read.** The `/physton_prompt/*_data*` routes (`storage.py:32`, `csv.py:43`, `styles.py:10`) resolve a caller-supplied key/type into a filesystem path with no containment → arbitrary write/read (plus a symlink escape in `styles`). Please confine these to the extension's data directory.

I'll re-scan once addressed.

ltdrdata:
Thanks for pushing the update — I re-cloned the repo at HEAD and re-ran the check. I want to be straight with you: the re-check still turns up the same set of items, so it looks like the new commit didn't reach these particular spots yet. Here's exactly what it's still seeing, so you know where to aim.

The root of it is that the node auto-starts the bundled web service on all interfaces when ComfyUI loads, and its routes have no access control — so any web page the user has open can reach them. That's what makes each of the items below reachable:

- Package install over an open route — sd-webui-prompt-all-in-one/scripts/on_app_started.py:88. POST /physton_prompt/install_package hands a caller-supplied package name straight to install_package(), which installs it. Please make this route local-only / user-initiated, or drop it, so a remote page can't trigger an install.

- Shell command built from the package name — sd-webui-prompt-all-in-one/scripts/physton_prompt/packages.py:42. install_package drops the package name into launch.run_pip, which builds a command string the shell then runs. Please pass the name as a separate argument instead of building a command string, so the value can't run as a command.

- File read by caller-supplied key — sd-webui-prompt-all-in-one/scripts/physton_prompt/csv.py:47. GET /physton_prompt/get_csv builds the served path as base_dir + key with no containment, so a key with ".." reads files outside that folder. Please confine it: resolve the final real path and check it stays inside the fixed base (realpath + commonpath), and reject absolute paths and "..".

- File write by caller-supplied key — sd-webui-prompt-all-in-one/scripts/physton_prompt/storage.py:78. POST /physton_prompt/set_data uses the caller's key directly in the filename with no containment, so a key can write outside the folder. Same fix: confine the final path to the fixed base and reject absolute paths and "..".

One thing worth flagging, since it's an easy trap: a loopback / same-machine check isn't enough on its own here. The request comes from the user's own machine — a page they have open makes it from there — so it looks local even when it isn't user-initiated. These actions need to be confined or made local-only / user-initiated rather than gated on where the request appears to come from.

---

Update — two more items from a fuller re-check, same root cause:
- POST /api/expand-prompt (app.py:948) takes the API base from the request and falls back to the server's LLM_API_KEY, so the stored key is sent to whatever host the request names. Please only use the stored key with its own configured endpoint, or restrict the host to an explicit allow-list.
- POST /physton_prompt/gen_openai and /physton_prompt/translate(s) (sd-webui-prompt-all-in-one/scripts/physton_prompt/gen_openai.py:9) take api_base / endpoints from the request body and return the response. Please restrict these to an explicit allow-list of hosts.

chenr5934-tech:
Sorry for the mess around this submission.

I opened three PRs for the same thing (#3096, #3097, and this one). #3096
deleted most of custom-node-list.json by accident — that was my mistake —
and the later ones ended up as duplicates of each other. I've closed this
one too so it stops sitting in the queue.

On the review itself: understood, and thanks for spelling it out in that much
detail. To be clear about what actually needs to change here — the install
route being reachable from a page, the shell command string built out of a
package name, and the two path-traversal reads and writes — all of those have
to go. The sd-webui-prompt-all-in-one code is something I ported and bundled
rather than wrote, but that doesn't make it your job to untangle.

I haven't fixed them yet. I'd rather come back with a real fix than an
explanation, so I won't re-submit until that's done.

One correction to my own thinking, since this is the part I had wrong: I had
assumed a same-machine check would be enough. Your note about the request
coming from the user's own machine is what made it click that it isn't.

PR #3113, opened 2026-07-28
Add ComfyUI-Visual-LoRA-Loader

DarthNeurad:
Adding Visual LoRA Loader to the custom node list.

### DIFF
--- a/custom-node-list.json
+++ b/custom-node-list.json
@@ -60909,6 +60909,16 @@
             ],
             "install_type": "git-clone",
             "description": "Cycle through the number input fields on a node using Tab / Shift+Tab. Classic UI QoL extension."
-        }
+        },
+		{
+			"author": "Neurad",
+			"title": "ComfyUI-Visual-LoRA-Loader",
+			"reference": "https://github.com/DarthNeurad/ComfyUI-Visual-LoRA-Loader",
+			"files": [
+			  "https://github.com/DarthNeurad/ComfyUI-Visual-LoRA-Loader"
+			],
+			"install_type": "git-clone",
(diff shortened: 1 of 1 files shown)


ltdrdata:
Thanks for the submission. I have to hold this — one route reads an attacker-controlled path with no containment.

**Arbitrary file read — `POST /neurad/fetch-single-info`** (`visual_lora_loader.py:260`): the `path` field from the request body is resolved with `os.path.join(root, path)` gated only by `os.path.exists` — no `realpath`+`commonpath` confinement, no `..`/absolute rejection — and the file's contents are returned in the response. A `../`-laden or absolute `path` reads any file on the host.

Please confine `path` to the intended cache/model root (`realpath`+`commonpath`, reject `..`/absolute). I'll re-scan once addressed.

DarthNeurad:
Hello,

Thank you for the detailed security audit. I have addressed the Arbitrary File Read vulnerability in visual_lora_loader.py (endpoint POST /neurad/fetch-single-info).

Changes Made:

    Implemented strict path confinement using os.path.realpath() and startswith() checks.
    Added explicit rejection of absolute paths.
    Added logging for blocked traversal attempts.

Verification:

    I have written and executed a local test suite confirming that directory traversal (../) and absolute path attacks now return 404/400 errors instead of file contents.
    Legitimate file access remains functional.

The fix has been pushed to the repository  (Commit: 0bd444e).

Please let me know if you are ready to proceed with the re-scan.

Best regards,

Neurad

ltdrdata:
Re-reviewed at the current HEAD. The arbitrary file read in `/neurad/fetch-single-info` is properly contained now (realpath + separator-guarded root check + absolute-path rejection), and the endpoint returns only hash-keyed metadata rather than file bytes. Registered to the manager node list. Thanks for the fix.

PR #2821, opened 2026-04-22
Add ComfyUI-Pixaroma to custom-node-list.json

pixaroma:
This PR adds the public custom node repository ComfyUI-Pixaroma by pixaroma to the Manager database. The change only updates the root custom-node-list.json and keeps the JSON valid.

PR #3118, opened 2026-08-01
Re-add ComfyUI-Pixaroma to custom-node-list.json

pixaroma:
The entry was merged in #2821 (May 2026) but is no longer in the list. The most likely reason is that my GitHub account was suspended for over a week, so the repository would have returned 404 to the scanner during that window. The account and the repository are fully restored.

Placed after the last git-clone block, matching the placement fix in 6ffead7b.

### DIFF
--- a/custom-node-list.json
+++ b/custom-node-list.json
@@ -61017,6 +61017,17 @@
             ],
             "install_type": "git-clone",
             "description": "Visual interface for LoRA management with tabs, drag-and-drop, and automatic CivitAI metadata fetching."
+        },
+        {
+            "author": "pixaroma",
+            "title": "ComfyUI-Pixaroma",
+            "id": "comfyui-pixaroma",
+            "reference": "https://github.com/pixaroma/ComfyUI-Pixaroma",
+            "files": [
+                "https://github.com/pixaroma/ComfyUI-Pixaroma"
+            ],
+            "install_type": "git-clone",
+            "description": "Creative suite for ComfyUI: fullscreen Image Composer, Paint, 3D Builder, Image Crop and AudioReact editors, plus Load/Save Image, Save Mp4, Load Video, Inpaint Crop and Stitch, Outpaint, XY Plot, Prompt library, LoRA Loader, Control Panel, Notes, Labels, and canvas tools for aligning and grouping nodes."
         }
     ]
 }


ltdrdata:
Thanks for the re-add. Before this can go back on the list, the pack has a few filesystem-access issues — a crafted `/prompt` (all node widgets are unauthenticated-reachable) can read and write arbitrary files on the host:

**1. Arbitrary file read (`nodes/node_audio_studio.py:47-96`).** `image_path`/`audio_path` come from the hidden `studio_json` widget and are joined as `PIXAROMA_INPUT_ROOT / rel_path` with no containment. `pathlib`'s `/` drops the base on an absolute right-hand side (`Path('/base')/'/etc/passwd' == Path('/etc/passwd')`), and `../` also escapes, so any file readable by the process is loaded via `PIL.Image.open()`/`wave.open()` and returned as a tensor (routable to a Save/Preview node). Every other loader in the pack already uses `realpath()`+`startswith(base+os.sep)` — apply the same guard here.

**2. Arbitrary file write (`nodes/_save_helpers.py:201`, `node_save_image.py`).** `_resolve_save_folder` returns any absolute `folder` verbatim; `inside_output` is computed but never enforced, so a crafted prompt writes a Pillow-encoded image to any writable path. Please enforce containment at write time, not just for the UI flag.

**3. Uncontained folder browse/read (`server_routes.py:1691/1725/1754`).** `/pixaroma/api/load_images_folder/{list,thumb,browse}` operate on a raw `?path=` with no root containment; `/thumb` returns image bytes for any file. Please constrain these to the ComfyUI input/output roots.

Also `/pixaroma/api/save_image/open_folder` (`server_routes.py:2024`) runs `xdg-open`/`os.startfile` on any absolute directory with no containment — the sibling `/workflows/reveal` route already checks `_is_path_under`, so mirror that.

Once the read/write paths are contained, happy to take another look.

pixaroma:
@ltdrdata 
Thanks for the detailed review. All four are fixed in v1.4.82, and I swept the pack for the same class rather than just the reported sites.

The four: AudioReact and Save Image now enforce containment through one shared guard; the three `load_images_folder` routes are rooted (note `/thumb`'s existing check validated against the caller's own `?path=`, so it proved nothing); `open_folder` mirrors `/workflows/reveal`, and since a UNC path leaks an NTLM hash during the `isdir` itself, UNC is judged lexically before anything resolves it.

Same pass also fixed: `next_counter` (same uncontained resolver, a read-oracle), the `IS_CHANGED` of four editor nodes, the Load Images node itself and not just its routes, and a missing dimension cap in `remove_bg`.

Rather than restricting to input/output, a folder is allowed if it's under ComfyUI's own directories or the user picked it in the native OS dialog. No endpoint writes the allowlist, and the dialog's start location must already be approved so a pre-selected path can't be one-click approved.

ltdrdata:
Verified against v1.4.82 — the four sites are contained now, the wider sweep checks out, and remember_folder stays on the picker route. I've added it back to the list directly, so I'm closing this PR (the branch had drifted into a conflict with main and the entry is in now).

Two small follow-ups for whenever you're next in there: `_resolve_inline` (node_audio_studio) and `save()` (node_save_image) still resolve before the lexical screen, so a `\\host\share` or `%HOMESHARE%` value reaches `realpath` before it's refused — `safe_join()` and `prescreen_folder_field()` respectively line them up with what the routes already do. The read/write itself is already refused either way.

Thanks for the thorough pass.

PR #3120, opened 2026-08-02
Add Multi Bool Panel to custom node registry

solidlime:
Adds the Multi Bool Panel custom node (https://github.com/solidlime/ComfyUI-MultiBoolPanel) to the registry.

- Single node with up to 20 named boolean toggle switches, each with an independent BOOLEAN output socket.
- Toggle count adjustable (slot_count 1-20), per-toggle labels, all-switch master.
- Batch operation: name_filter + broadcast_mode (widgets/bypass/mute) applies to matching nodes by name.


ltdrdata:
Reviewed the pack — pure frontend graph toggle helper, no backend/file/network surface. Registered to the manager node list.

PR #3122, opened 2026-08-03
Add AtlasCloud_ComfyUI to custom-node-list.json

binyangzhu000-sudo:
Registers `AtlasCloudAI/atlascloud_comfyui` in `custom-node-list.json`.

The node pack calls Atlas Cloud's hosted inference API, so the nodes run without local model weights or a GPU: text-to-image, image edit, text-to-video and image-to-video (Seedance 2.0, Kling 3, Sora 2, Veo 3.1, Wan 2.7, Nano Banana 2/Pro, GPT Image 2, Flux 2, Seedream 5), plus LLM chat and TTS nodes. An Atlas Cloud API key is required.

Repo: https://github.com/AtlasCloudAI/atlascloud_comfyui
`install_type` is `git-clone`; `requirements.txt` is present for automatic dependency install.

JSON validated with `python -m json.tool` after the change (single entry appended, no other lines touched).

### DIFF
--- a/custom-node-list.json
+++ b/custom-node-list.json
@@ -60978,6 +60978,17 @@
             ],
             "install_type": "git-clone",
             "description": "Creative suite for ComfyUI: fullscreen Image Composer, Paint, 3D Builder, Image Crop and AudioReact editors, plus Load/Save Image, Save Mp4, Load Video, Inpaint Crop and Stitch, Outpaint, XY Plot, Prompt library, LoRA Loader, Control Panel, Notes, Labels, and canvas tools for aligning and grouping nodes."
+        },
+        {
+            "author": "AtlasCloudAI",
+            "title": "AtlasCloud_ComfyUI",
+            "id": "atlascloud_comfyui",
+            "reference": "https://github.com/AtlasCloudAI/atlascloud_comfyui",
+            "files": [
+                "https://github.com/AtlasCloudAI/atlascloud_comfyui"
+            ],
+            "install_type": "git-clone",
+            "description": "Nodes for running Atlas Cloud's hosted AI models inside ComfyUI without a local GPU or model weights: text-to-image, image edit, text-to-video and image-to-video via Seedance 2.0, Kling 3, Sora 2, Veo 3.1, Wan 2.7, Nano Banana 2/Pro, GPT Image 2, Flux 2 and Seedream 5, plus LLM chat and TTS nodes. Requires an Atlas Cloud API key."
         }
     ]
 }


ltdrdata:
Path containment in these routes is broken. `asset_path`/`input_asset_path` check with `relative_to(assets_dir / prediction_id)` — a base built from the caller's own input, which defeats the check — and `_run_path` has no containment at all; the `{prediction_id}` segment also accepts encoded separators. As a result a remote caller can read files outside the intended directory, and `POST …/refresh-assets` builds its write paths the same way, so it can also write outside the store.

Please contain every path with `realpath()` + `commonpath([real, base]) == base` against a fixed base, and reject any `prediction_id`/`filename` containing a path separator. These are open routes with no access control, so any remote caller can call them; please also make `/refresh-assets` local-only / explicitly user-initiated, or remove the route.

Separately, the node UI strings (tooltips under `src/atlascloud_comfyui/nodes/`, labels in `web/js/atlas_history_panel.js`) are in Chinese — please use English. Locale feature: https://github.com/Comfy-Org/ComfyUI/pull/6558

binyangzhu000-sudo:
@ltdrdata thanks — you were right, and it reproduced exactly as you described. Fixed in AtlasCloudAI/atlascloud_comfyui#111.

**Path containment.** The check was `relative_to(assets_dir / prediction_id)`, and since the base came from the caller's own route parameter it moved with the input and passed for any traversal. Confirmed before fixing:

```
asset_path('../../outside', 'config.json')
  → /tmp/.../outside/config.json     # outside the history store
```

`_run_path` and the per-run dirs had no containment at all, and those feed writes through `record_submission` and `refresh_downloaded_assets` — so the write side you flagged was real too.

Now every path goes through `realpath()` + `commonpath([real, base]) == base` against a **fixed** root, and `prediction_id`/`filename` are rejected if they contain `/`, `\`, `%2f`, `%5c` or NUL, or are `.`/`..`. Route handlers return 400 on a bad segment rather than relying on a `None` further down.

**Auth on `/refresh-assets`.** Agreed it shouldn't be open. I didn't invent a token scheme — ComfyUI has none to reuse, and its own protection for state-changing routes is `create_origin_only_middleware` + `is_loopback` in `server.py`, which is skipped entirely under `--enable-cors-header`. So the route now applies that same loopback rule directly, rather than depending on a middleware that may be disabled. If you'd rather it hook into something else, tell me what and I'll switch.

**Verification** — a regression test covering read and write paths (including encoded separators) fails **7 cases** against the previous implementation and passes on the fix; full suite **712 passed, 17 skipped**.

**English UI strings** — acknowledged, not done yet: 121 lines across 32 files, and I'd rather not bury a security fix under a localization diff. Doing it as a separate pass, and I'll use the locale feature you linked.

This PR here only adds the `custom-node-list.json` entry, so nothing in it changes. Happy for you to hold it until the fixes land on our side — say the word if you'd prefer I close and re-open after.

ltdrdata:
Thanks for the quick turnaround, and for reproducing it so precisely — that write-up is exactly right. I re-cloned at HEAD and re-checked.

The asset download path is genuinely confined now: routing every path through realpath() plus a commonpath check against a fixed base, and rejecting prediction_id/filename that contain slashes, %2f, %5c, NUL, or are "." / "..", closes the traversal on that side. Tightening the per-run directories and _run_path the same way was the right call too — that write side was real. Nice work, and thanks for the regression test.

Two write paths still fall outside that confinement, though — the preview nodes build their save path the same way the download side used to:

- src/atlascloud_comfyui/nodes/utils/video_previewer.py:78 — the subfolder and filename_prefix widget values are only stripped of leading/trailing slashes, not "..", so the joined path can climb out of the output directory and the downloaded bytes land at an arbitrary location. Send this path through the same fixed-base check you added downstream (realpath + commonpath against the output root), and reject any segment that is absolute or contains "..".
- src/atlascloud_comfyui/nodes/utils/image_preview.py:109 — same shape: subfolder and filename_prefix aren't checked for "..", so os.makedirs and the image save can write outside the output directory. Same remedy — confine the joined path to the output root before makedirs and before saving.

Separately, the history panel UI is still coming through in Chinese on my scan. I saw your note that the English strings are a later pass with the locale feature — that's fine to land separately, just flagging that the re-scan still sees them.

---

Correction — I credited _run_path above, but that was my mistake: _run_path (history/local_history.py:325-326) still builds runs_dir / f"{prediction_id}.json" with no containment, and GET /atlas/history/runs/{prediction_id} (history/api.py:49-57) returns what it reads. Please route it through the same realpath + commonpath check and prediction_id validation you added for the asset paths.

PR #3132, opened 2026-08-05
Add ComfyUI_Eagle_Suite to custom-node-list.json

Yao3596:
This PR adds **ComfyUI_Eagle_Suite** (https://github.com/Yao3596/ComfyUI_Eagle_Suite) to the ComfyUI-Manager custom node list.

It provides Eagle image gallery browsing, LoRA gallery with Civitai integration, Eagle image saver with metadata/tags, and multi-profile LLM API loader nodes.

### DIFF
--- a/custom-node-list.json
+++ b/custom-node-list.json
@@ -45001,7 +45001,7 @@
             "install_type": "git-clone",
             "description": "Prompt Style Selector with Preview, Multiple Selecting, Search and Local Base."
         },
-		{
+        {
             "author": "DemonAlone",
             "title": "DemonAlone-JS_Addon-ComfyUI",
             "reference": "https://github.com/DemonAlone/DemonAlone-JS_Addon-ComfyUI",
@@ -60926,7 +60926,10 @@
             ],
             "install_type": "git-clone",
             "description": "Professional ACES and EXR workflow nodes with OCIO 2.1 support and sequence loading.",
-            "pip": ["opencolorio>=2.2.0", "openexr>=3.2.0"]
+            "pip": [
+                "opencolorio>=2.2.0",
+                "openexr>=3.2.0"
+            ]
         },
         {
             "author": "Neurad",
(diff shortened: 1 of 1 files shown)


ltdrdata:
The Eagle routes are open (no access control) and take arbitrary host paths:
- `read_file` (lora_browser.py:62) → returns any file's contents.
- `deleteImage`/`deleteLora` (image_browser.py:221, lora_browser.py:45) → `os.remove` any path.
- rename (image_browser.py:277, lora_browser.py:130, audio_browser.py:81) → `os.path.join(dirname, new_name)`, so an absolute `new_name` = arbitrary move.
- `upload` (image_browser.py:92) + rename → write to any path.
- `GET /eagle/{file_type}/{path:.*}` (image_browser.py:119) → no `..` check.

`is_safe_path()` isn't wired to these (needs `EAGLE_STRICT_PATHS`). Please confine every file op to an allowed base (`realpath`+`commonpath`, reject absolute/`..`). I'll re-review once fixed.

ltdrdata:
Thanks! A few things to fix before this can be registered:

1. Several save nodes write files / create dirs from a free-text directory widget with no containment (arbitrary-path write): eagle_saver.py:167, text_nodes.py:76, video_nodes.py:470 (is_safe_path imported but never called), advanced_video_saver.py:210. Confine each save directory with `realpath` + `commonpath` (reject absolute / `..`), not just the filename.
2. The open `update_config` (prompt_presets.py:830) and `authorize_media_root` (tools_utils.py:220) let a caller set the allowed root to `/`, exposing arbitrary file read via `/cover` and `/EagleAudioList/stream`. Confine these roots to a fixed base and don't let a route widen them.
3. Two routes fetch a caller-supplied host and return the response with only a denylist: prompt_presets.py:899 (obsidian.api_url), api_key_node.py:269 (base_url). Restrict the outbound host to an allow-list.
4. local_llm_node.py:621 uses `trust_remote_code=True` on a widget-named model path — with the write nodes above, an attacker can plant and load a model to run code. Remove `trust_remote_code=True` or pin the path to a fixed location.
5. UI strings are in Chinese (display names, CATEGORY, RETURN_NAMES, tooltips, combo labels, JS text). Write them in English — locale feature: https://github.com/Comfy-Org/ComfyUI/pull/6558

---

Update — one more item from a fuller re-check: POST /danbooru_search/settings (eagle_suite/danbooru_search.py:1407) saves api_base_url from the request, and _is_allowed_image_url (danbooru_search.py:354-368) then allows whatever host that is, so /danbooru_search/image_proxy fetches and returns content from any https host, and the Danbooru API calls (which can carry the stored danbooru_api_key) go to it too. Please restrict api_base_url to an explicit allow-list of hosts (e.g. danbooru.donmai.us) and make the settings route local-only / user-initiated.

PR #3136, opened 2026-08-06
fix: report the real reason a batch install or uninstall is denied

### LINKED ISSUE #3128: Batch install/uninstall map every 404 to the "default channel" message, masking the real denial reason
**Version:** ComfyUI-Manager 3.41, commit `d404e623` (clean `main`, unmodified checkout)

## Summary

Two denial-diagnostics defects that compound. A user blocked by the loopback term of `allow_git_url_install` is told their **channel / security level** is the problem, and the log line the dialog defers to tells them to enable a setting that is **already enabled**. Following both messages exactly leaves you stuck, with no indication of the actual gate.

This is about the messaging only — the loopback bound itself is intentional per #2991.

## 1. `/manager/queue/install` discards the server's 404 body

The server returns a distinguishable body (`glob/manager_server.py:1498` and `:1503`):

```python
return web.Response(status=404, text="A security error has occurred. Please check the terminal logs")
```

The client throws it away (`js/custom-nodes-manager.js:1484`):

```js
} else if(res.status == 404) {
    errorMsg += `With the current security level configuration, only custom nodes from the <B>"default channel"</B> can be installed.\n`;
} else {
    errorMsg += await res.text() + '\n';
}
```

The `else` branch immediately below already does the right thing.

The substituted copy points at `security_level` and channel configuration, and **neither can resolve this denial**. The `risky_level == 'high'` arm routes through `is_dedicated_install_allowed()` and never consults `security_level` at all — setting `security_level = weak` changes nothing. That is a false lead users do follow.

The uninstall path has the same defect at `js/common.js:667`.

For contrast, the dedicated `/customnode/install/git_url` surface handles this correctly: 403, with `js/common.js:274` naming the responsible flag. Only the batch path is affected.

## 2. `SECURITY_MESSAGE_FLAG_GIT_URL` omits the loopback requirement

`glob/manager_server.py:40`:

> ERROR: This action requires 'allow_git_url_install = true' in config.ini ([default] section). This setting is independent of security_level.

But the predicate is flag **AND** loopback (`glob/manager_server.py:90`):

```python
def is_dedicated_install_allowed(flag_value: bool, listen_address: str) -> bool:
    return bool(flag_value) and is_loopback(listen_address)
```

With `allow_git_url_install = True` and `--listen 0.0.0.0`, that message is emitted verbatim — instructing the user to set a value that is already `true`, and never mentioning the network-position requirement. `SECURITY_MESSAGE_FLAG_PIP` has the same omission. By comparison, `SECURITY_MESSAGE_NORMAL_MINUS` does spell out its `--listen` condition.

I suspect this accounts for a fair share of the traffic on #3023: many reports there are of the form "I set it to true and it still doesn't work", which is exactly what this message produces on a non-loopback listener.

## Reproduction

`user/__manager/config.ini`:

```ini
[default]
security_level = normal
allow_git_url_install = True
```

Launch: `python main.py --listen 0.0.0.0 --port 8188`

Install any node pack whose repo URL is not in `custom-node-list.json`, from the Custom Nodes Manager UI.

**Actual**

- Dialog: `[Installation Errors] '<pack>': With the current security level configuration, only custom nodes from the "default channel" can be installed.`
- Log: `ERROR: This action requires 'allow_git_url_install = true' in config.ini ([default] section).` — while it is already `True`.

**Expected**

A message naming the gate that actually fired, e.g. *"Install via git URL requires `allow_git_url_install = true` AND ComfyUI listening on a loopback address (currently: 0.0.0.0)."*

## Suggested fix

- `js/custom-nodes-manager.js:1484` and `js/common.js:667`: surface `await res.text()` for 404 as the `else` branch does, or give the two 404 arms distinct copy.
- `SECURITY_MESSAGE_FLAG_GIT_URL` / `SECURITY_MESSAGE_FLAG_PIP`: state the loopback requirement alongside the flag, ideally including the current `--listen` value.

Dhevenddra:
Fixes #3128.

## Problem

**1. The batch install path throws away the server's 404 body.**

`js/custom-nodes-manager.js` replaced every 404 with a message about the `"default channel"` and the security level. The denial it fires on most often cannot be resolved by either. In `install_custom_node`, the `risky_level == 'high'` arm goes through `is_dedicated_install_allowed(core.get_config()['allow_git_url_install'], args.listen)`, which is `flag AND loopback` and never reads `security_level`. Setting `security_level = weak` changes nothing, so the message is a false lead that users do follow.

The same branch also swallowed the informative 404s on that route, for example ``Following node pack doesn't provide `nightly` version``, which the user never got to see.

**2. The uninstall path carried the same substitution for a status it never returns.**

`uninstall_custom_node` returns 403 or 200 only. The 404 arm in `js/common.js` could therefore only be reached when the endpoint itself was missing (an outdated or partly-installed Manager), where a note about channel configuration points in exactly the wrong direction.

**3. `SECURITY_MESSAGE_FLAG_GIT_URL` omits half the gate.**

The message asks for `allow_git_url_install = true` and adds that the setting is independent of `security_level`, but `is_dedicated_install_allowed` also requires a loopback listen address. Anyone on `--listen 0.0.0.0` is told to enable a flag they already enabled, with no hint about the real bound. `SECURITY_MESSAGE_FLAG_PIP` has the identical gap.

## Fix

- Drop the 404 special case in both JS paths so the existing `else` arm surfaces the server's own body. This is what the 403 path already does, and the dedicated `/customnode/install/git_url` surface already names the responsible flag through `js/common.js`.
- Name the loopback requirement in both flag messages.

The loopback bound itself is deliberate per #2991 and is unchanged here. This is diagnostics only: no gate, status code, or policy moves.

## Evidence

`python -m pytest tests -q` on a clean 3.12 venv:

```
38 passed, 14 skipped, 22 subtests passed in 36.09s
```

`ruff check .` (the CI gate) passes.

The denial-copy guard in `tests/test_install_flags_gates.py` now asserts the loopback requirement. Reverting only `glob/manager_server.py` and re-running it fails as expected:

```
E   AssertionError: 'loopback' not found in "ERROR: This action requires
    'allow_git_url_install = true' in config.ini ([default] section).
    This setting is independent of security_level. ..."
    : constant must name the loopback requirement
FAILED tests/test_install_flags_gates.py::DenialConstantsTest::test_flag_constants_content
```

The JS side has no test harness in this repo, so those two changes are deletions that fall through to the existing `else` branch rather than new logic.

### DIFF
--- a/glob/manager_server.py
+++ b/glob/manager_server.py
@@ -37,8 +37,8 @@ SECURITY_MESSAGE_MIDDLE_OR_BELOW = "ERROR: To use this action, a security_level 
 SECURITY_MESSAGE_NORMAL_MINUS = "ERROR: To use this feature, you must either set '--listen' to a local IP and set the security level to 'normal-' or lower, or set the security level to 'middle' or 'weak'. Please contact the administrator.\nReference: https://github.com/ltdrdata/ComfyUI-Manager#security-policy"
 SECURITY_MESSAGE_GENERAL = "ERROR: This installation is not allowed in this security_level. Please contact the administrator.\nReference: https://github.com/ltdrdata/ComfyUI-Manager#security-policy"
 SECURITY_MESSAGE_NORMAL_MINUS_MODEL = "ERROR: Downloading models that are not in '.safetensors' format is only allowed for models registered in the 'default' channel at this security level. If you want to download this model, set the security level to 'normal-' or lower."
-SECURITY_MESSAGE_FLAG_GIT_URL = "ERROR: This action requires 'allow_git_url_install = true' in config.ini ([default] section). This setting is independent of security_level. Reference: https://github.com/ltdrdata/ComfyUI-Manager#security-policy"
-SECURITY_MESSAGE_FLAG_PIP = "ERROR: This action requires 'allow_pip_install = true' in config.ini ([default] section). This setting is independent of security_level. Reference: https://github.com/ltdrdata/ComfyUI-Manager#security-policy"
+SECURITY_MESSAGE_FLAG_GIT_URL = "ERROR: This action requires 'allow_git_url_install = true' in config.ini ([default] section) AND ComfyUI listening on a loopback address (e.g. --listen 127.0.0.1). The flag alone has no effect when ComfyUI listens on any other address. This setting is independent of security_level. Reference: https://github.com/ltdrdata/ComfyUI-Manager#security-policy"
+SECURITY_MESSAGE_FLAG_PIP = "ERROR: This action requires 'allow_pip_install = true' in config.ini ([default] section) AND ComfyUI listening on a loopback address (e.g. --listen 127.0.0.1). The flag alone has no effect when ComfyUI listens on any other address. This setting is independent of security_level. Reference: https://github.com/ltdrdata/ComfyUI-Manager#security-policy"
 
 routes = PromptServer.instance.routes
 
--- a/js/common.js
+++ b/js/common.js
@@ -663,8 +663,6 @@ export async function uninstallNodes(nodeList, options = {}) {
 
 			if (res.status == 403) {
 				errorMsg += `This action is not allowed with this security level configuration.\n`;
-			} else if (res.status == 404) {
-				errorMsg += `With the current security level configuration, only custom nodes from the <B>"default channel"</B> can be uninstalled.\n`;
 			} else {
 				errorMsg += await res.text() + '\n';
 			}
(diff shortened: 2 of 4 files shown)


Dhevenddra:
Closing this as superseded. ffb2f03, 4fe65b0 and 98be899 landed after I opened it and cover the same ground more thoroughly than this branch did.

For anyone reading back: the 404 change here is the same edit as 4fe65b0, and the constants rewrite in ffb2f03 goes further than mine by printing the live `--listen` value at each emission site, spelling out that both values are read once at startup, and adding a dedicated message for the blocked-risk arm instead of reusing the general one.

The one piece not covered upstream is a regression guard, since `_assert_honest_copy` pins the flag name and `config.ini` but not the loopback half of the predicate. I have opened that separately as #3216.

PR #3137, opened 2026-08-06
Register muapi-comfyui in the custom node list

Anil-matcha:
## Summary

- Register the existing SamurAIGPT/muapi-comfyui package in ComfyUI-Manager’s custom node list.
- Include its current image and video model coverage and API-key requirement.
- Keep installation pointed at the public package repository.

Validated that custom-node-list.json remains valid JSON.

### DIFF
--- a/custom-node-list.json
+++ b/custom-node-list.json
@@ -60988,6 +60988,16 @@
             ],
             "install_type": "git-clone",
             "description": "Multi-point white balance and exposure correction for e-commerce product images."
+        },
+        {
+            "author": "SamurAIGPT",
+            "title": "muapi-comfyui",
+            "reference": "https://github.com/SamurAIGPT/muapi-comfyui",
+            "files": [
+                "https://github.com/SamurAIGPT/muapi-comfyui"
+            ],
+            "install_type": "git-clone",
+            "description": "ComfyUI custom nodes for AI image and video generation through MuAPI, including Seedance, Kling, Veo3, Flux, HiDream, GPT-image, and Imagen4. Requires a MuAPI API key from https://muapi.ai."
         }
     ]
 }


ltdrdata:
1. The `save_subfolder` / `filename_prefix` widgets build a write path with no confinement, so files can be written outside the intended directory.
   - muapi_video_saver_node.py:40-53
2. A `video_url` widget is fetched with only an `http` prefix check (no host allow-list).

Please confine the write path to a fixed base directory and restrict the fetch host. I'll re-evaluate once both are constrained.

Anil-matcha:
@ltdrdata Thanks for identifying these security issues. I opened [MuAPI ComfyUI PR #2](https://github.com/SamurAIGPT/muapi-comfyui/pull/2) with commit `b39b8f9`.

- `save_subfolder` is now resolved and confined beneath ComfyUI's output directory.
- `filename_prefix` is restricted to one filename component.
- `video_url` now requires HTTPS and the exact `cdn.muapi.ai` host.
- HTTP redirects are disabled so the request cannot silently move to an unapproved host.
- Seven offline regression tests cover traversal, arbitrary hosts, redirects, and the valid save path.

Ruff, Python compilation, and `git diff --check` also pass. No live or billable MuAPI request was made.

Please re-evaluate [ComfyUI Manager #3137](https://github.com/Comfy-Org/ComfyUI-Manager/pull/3137) after the security fix is reviewed.

Anil-matcha:
@ltdrdata The security fix has now been merged in [MuAPI ComfyUI #2](https://github.com/SamurAIGPT/muapi-comfyui/pull/2) at merge commit `5677bd8`.

The `main` branch now contains the confined output-path handling, single-component filename validation, HTTPS `cdn.muapi.ai` allow-list, and redirect blocking described above. The seven offline regression tests and static checks passed before merge.

Please re-evaluate [ComfyUI Manager #3137](https://github.com/Comfy-Org/ComfyUI-Manager/pull/3137) against the updated node repository.

ltdrdata:
Thanks for turning this around so quickly, and for adding the regression tests — that
helps a lot.

The save/download side landed cleanly, and I can confirm it on the re-check: the save
path is resolved and kept under the output directory, video_url is pinned to HTTPS on
the exact cdn.muapi.ai host, and redirects are off. Nice work on that part.

The read side is still open, though. _upload_file takes whatever local path it's handed
and sends the file's bytes to muapi.ai, so an arbitrary local file can still be read and
uploaded. It shows up in a few places:

- _upload_file (muapi_nodes.py:337) opens a caller-supplied path with no base-directory
  check before sending it — confine the path here: resolve it with realpath and require
  it to sit under a fixed base (the input/output directory) via commonpath, and reject
  absolute paths and "..".
- MuAPIGenerate (muapi_nodes.py:879) passes file_path_1/file_path_2 straight into
  _upload_file with only an existence check — apply the same confinement before the call.
- MuAPIImageToVideo (muapi_nodes.py:637) passes video_file_1..3 and audio_file_1..3 the
  same way — same confinement.
- MuAPILipsync (muapi_nodes.py:800) passes audio_file_path with only an existence check —
  same confinement.

So the save side is fixed; the read side at _upload_file still lets any local file be
read and sent off-host. Confining these paths the same way you already confined the save
path should close it.

Anil-matcha:
Thanks for the follow-up. I opened [MuAPI ComfyUI #3](https://github.com/SamurAIGPT/muapi-comfyui/pull/3) to address the remaining read-side issue.

The patch resolves user-supplied paths with `realpath`, rejects absolute paths and `..` components, requires the resolved path to remain under ComfyUI's input/output directories via `commonpath`, and applies the check before the generic, omni-reference image-to-video, and lip-sync upload paths can open a file. It includes offline regression coverage for traversal, absolute paths, symlink escapes, valid input/output paths, and the no-network boundary.

Once #3 is merged, please re-evaluate this registry PR against the updated node repository. No live or billable MuAPI request was made.

ltdrdata:
Thanks for the fix! I re-checked the repo after your PR #3 landed — local file reads are now confined to the ComfyUI input/output folders, and the download side uses a fixed host. I have registered muapi-comfyui directly in `custom-node-list.json` (the list tail had drifted so this PR could not merge cleanly). Closing this so the entry is not added twice; it is now available in ComfyUI-Manager.

PR #3139, opened 2026-08-06
Add ComfyUI-ModelPathManager to custom node list

lztango:
Adding my custom node **ComfyUI-ModelPathManager** to the registry.

**Features:**
- Dynamically add model search paths at runtime (no restart needed)
- Auto-deduplicate paths pointing to the same physical directory
- **Junction-based path merge**: solves duplicate model storage for plugins that hardcode `os.path.join(folder_paths.models_dir, "xxx")` paths, without copying files (Windows Junction, no admin rights needed)
- Automatic startup check for mergeable directories (dry-run report only)
- Web UI panel with scan/apply buttons (Ctrl+Shift+M)

**Repo:** https://github.com/lztango/ComfyUI-ModelPathManager
**License:** MIT

Verified: JSON syntax valid, list loads without issues (5880 entries).

### DIFF
--- a/custom-node-list.json
+++ b/custom-node-list.json
@@ -2723,6 +2723,16 @@
             "description": "Manage models: browsing, download and delete."
         },
         {
+            "author": "lztango",
+            "title": "ComfyUI-ModelPathManager",
+            "reference": "https://github.com/lztango/ComfyUI-ModelPathManager",
+            "files": [
+                "https://github.com/lztango/ComfyUI-ModelPathManager"
+            ],
+            "install_type": "git-clone",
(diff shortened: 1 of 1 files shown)


ltdrdata:
`POST /model-path-manager/api/merge/apply` is an open route (no access control), so any remote caller can invoke it. `set_main_models_path` (path_merge.py:120) accepts any dir with only an `isdir` check, then `_apply_merge`/`_apply_migrate` `shutil.move` your real `models/` files there and delete the originals — a remote relocate/wipe of model files. `browse_directory`/`add_config` similarly expose the filesystem. Please confine all paths to the ComfyUI tree (`realpath`+`commonpath`) and gate move/delete behind explicit local confirmation. I'll re-review once fixed.

PR #3146, opened 2026-08-10
Register MiniMax H3 Contex Loop

ethanfel:
Registers https://github.com/ethanfel/ComfyUI-MiniMaxH3-Contex-Loop in the default Manager custom-node list.\n\n- Package ID: comfyui-minimaxh3-contex-loop\n- Publisher: ethanfel\n- Public repository with pyproject.toml metadata\n- comfy node validate passes

### DIFF
--- a/custom-node-list.json
+++ b/custom-node-list.json
@@ -61029,6 +61029,17 @@
             ],
             "install_type": "git-clone",
             "description": "OpenAI API based prompt rewriting agents for MiniMax H3 T2VA/I2VA/L2VA/FL2VA and Ref2VA workflows, using the official H3 skill files."
+        },
+        {
+            "author": "ethanfel",
+            "title": "MiniMax H3 Contex Loop",
+            "id": "comfyui-minimaxh3-contex-loop",
+            "reference": "https://github.com/ethanfel/ComfyUI-MiniMaxH3-Contex-Loop",
+            "files": [
+                "https://github.com/ethanfel/ComfyUI-MiniMaxH3-Contex-Loop"
+            ],
+            "install_type": "git-clone",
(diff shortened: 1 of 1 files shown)


ltdrdata:
The `prompt-optimize` route fetches a URL taken from the request body with only a scheme check and returns the response, so a caller can make the server request an arbitrary host and read the reply.
- prompt_optimizer.py:305-310

There are also open routes that read files by absolute path and open a folder. Please restrict the fetch to a host allow-list, confine the file reads to a fixed base directory, and make the folder-opening local-only. I'll re-evaluate once constrained.

ethanfel:
@ltdrdata The reported routes have now been constrained in https://github.com/ethanfel/ComfyUI-MiniMaxH3-Contex-Loop/commit/212486fc2029cd60845cfd666a0bf89cf65dfe96:

- Prompt Optimizer uses a server-owned exact-origin allowlist and validates every redirect. Cross-origin redirects are rejected.
- Browser-facing file reads are confined to the configured ComfyUI input/output/temp roots, including symlink-escape protection.
- Arbitrary server-path asset imports were removed.
- Host-folder opening is restricted to loopback clients and ignores forwarded-address headers.
- Security regression tests were added and the runnable Python and JavaScript suites pass on both main and nightly.

Could you please re-evaluate the node for approval?

ltdrdata:
Thanks for the contribution! I've registered `MiniMax H3 Contex Loop` directly in `custom-node-list.json` — the list tail had drifted from your branch's base, so the PR couldn't be merged cleanly. Closing this so the entry isn't added twice; it's now available in ComfyUI-Manager.

Issue #213, opened 2023-12-04
IMPORT FAILED


NanThu2307:
Import Failed Issue with ComfyUI in, How do I fix it?

![Import Failed](https://github.com/ltdrdata/ComfyUI-Manager/assets/141979157/b0859e44-ec98-429a-b29d-65593a9f0297)

ltdrdata:
Issues related to the import fail should be addressed to the repository of each custom node developer. You can click on the title to go to the respective repository.

Issue #211, opened 2023-12-03
Support for Colab Users


xliry:
Hi,

Im using colab version of comfyui manager, few features dont work. it will be great if you can fix bugs and i can keep update this title

-install missing custom nodes feature dont work.

ltdrdata:
Can you provide specific details about the situation where it's not working?

xliry:
> Can you provide specific details about the situation where it's not working?

for example when there is missing node
![github-1](https://github.com/ltdrdata/ComfyUI-Manager/assets/76709233/9e351738-70ff-48b5-bfcc-77c1415e158b)

![github-2](https://github.com/ltdrdata/ComfyUI-Manager/assets/76709233/33693194-137c-471f-b994-229c1bbbe02a)
install missing custom nodes feature tab dont show missing notes,

![github-3](https://github.com/ltdrdata/ComfyUI-Manager/assets/76709233/eacd5acb-1b03-4305-b7fd-bb6ea093d2a7)
when i click install models tab is not opened

![github-4](https://github.com/ltdrdata/ComfyUI-Manager/assets/76709233/d2f06d96-af2f-40b4-844f-ff46fefc653e)
install custom nodes tab only show 27 node,

ltdrdata:
> > Can you provide specific details about the situation where it's not working?
> 
[...]

PR #3147, opened 2026-08-10
Add VALYDA Protocol

RebellionPictures:


RebellionPictures:
Following up on this one.

VALYDA is also published to the Comfy Registry, but both published versions
have been flagged with no reason given — status_detail is empty in the API and
the publisher view shows nothing either. Two support requests in Discord #help
have gone unanswered, and I've run Bandit against the published archive myself:
zero high, zero medium, 14 low (all try/except/pass and try/except/continue).

So this list entry is currently the only route by which anyone can install the
node from the Manager at default security settings.

The diff is a single entry, +12 −1, and all checks pass. Any chance of a look?

RebellionPictures:
Update: the merge conflict is resolved — upstream merged in as ed61b31, and
CodeRabbit now reports "merge-ready after normal checks and review". Nothing
outstanding from my side.

ltdrdata:
The `Records_Folder` widget value is used as a write destination with no confinement (absolute and `..` paths are accepted), so a file can be written to any directory the server can reach.
- nodes.py:1655 (resolved at nodes.py:146)

Please confine the output path to a fixed base directory. I'll re-evaluate once contained.

Additionally, please write the node UI strings in English (currently the DESCRIPTION, widget dropdown options, tooltips, and JS field labels are in German).

RebellionPictures:
Thank you — that was a real defect and a fair catch.

Fixed in 2.3.2, published to the registry just now. Records_Folder is no longer
used as a write destination directly. It is resolved by a single function that
treats the entry as a relative sub-path under a fixed base of output/valyda,
and:

- rejects drive letters and UNC paths, tested with ntpath as well as os.path so
  "C:\..." and "\\server\share" are refused on every platform
- rejects a leading "/" or "\"
- rejects isabs under os.path, ntpath and posixpath
- rejects any ".." segment, leading or embedded
- and then, after the path is fully resolved with normpath, applies a final
  containment check using realpath plus commonpath against the fixed base —
  which also catches a symlink inside the base pointing outward

The same containment was added to the other function that builds a write
folder, deliberately checked against the fixed base rather than against a
caller-supplied one, so the guarantee does not depend on the caller.

Only valyda_protokoll/nodes.py changed between 2.3.1 and 2.3.2.

I see 2.3.0 has been banned — understood, and correct given the finding. 2.3.1
carries the same defect and should be treated the same way. 2.3.2 is the only
version that should be considered.

RebellionPictures:
The English interface is done, published as 2.3.3.

What is English now: the DESCRIPTION on all three node classes, every input
tooltip, the field labels and banner text in web/valyda.js, the server-side
info lines in the Project node's records box, and the output names. The twelve
hard-coded language branches in the JS are gone — all three nodes are
monolingual English.

I also added a locales/ folder following ComfyUI's own layout — en/ and de/,
each with main.json and nodeDefs.json, every key validated against an existing
node, input, value or output. en matches the source word for word. (German is
not currently among the interface languages the frontend ships, so locales/de
is prepared rather than active; that is noted in locales/README.md.)

One exception, which I want to be upfront about rather than quietly leave:
the dropdown option values are still bilingual, for example "both / beide".

Two reasons:

1. ComfyUI writes the chosen value into the saved workflow and rejects it on
   the next run if it is no longer in the list ("Value not in list"). Fourteen
   lists are affected. In two of them the value is additionally written verbatim
   into the generated disclosure document — the broadcaster field and three of
   the publication-channel entries. Renaming them would break every saved
   workflow containing such a value, and break the correspondence between the
   code and records users have already delivered to broadcasters. For a
   compliance tool I did not think that was an acceptable trade.

2. As far as I can tell the locale feature does not cover this case: the
   frontend does not read the options key for widget values. I could not find
   a lookup for it in the shipped bundle, and the bundled locale files contain
   no such entries. Display name, description, input names, tooltips and output
   names are translated; dropdown values are not. If I have missed something
   there, I would genuinely like to know — the mapping tables are already in
   place in both locales/en and locales/de, so the moment the frontend supports
   value translation the dropdowns become English with no code change.

For what it is worth on the verification side: 399 checks across twelve
scripts, and the 31 generated documents are textually identical before and
after the change once timestamps, run identifiers and checksums are factored
out. No function, no data-format field name and no stored value was altered —
display strings only. A new script freezes every dropdown list by count and
SHA-256 and fails if an interface string contains German.

So both of your points are now addressed: the path confinement in 2.3.2, the
English interface in 2.3.3. 2.3.3 is the version to look at.

RebellionPictures:
A status note, and one question about process.

Both of your points are addressed and published:

  2.3.2 — Records_Folder confined to a fixed base directory
  2.3.3 — English node interface, plus a locales/ folder

The registry has since flagged both of them. Current state:

  2.3.0  Banned
  2.3.1  Banned
  2.3.2  Flagged
  2.3.3  Flagged

status_detail is empty on all of them, and the publisher view shows no reason,
so I cannot tell whether 2.3.3 was flagged for something specific or whether
the scan simply tags this class of node — it writes files into the ComfyUI
output folder and registers three read-only GET routes, both of which appear
to be tagged capabilities rather than defects. Three other publishers describe
the same pattern in Comfy-Org/registry-backend#207, #209, #211 and #213.

The question: is releasing a flagged version your call, or does that sit with
someone else on the registry side? I ask because if the scan will not pass this
node regardless of what I change, then I am waiting on a review that no code
change can trigger, and I would rather know that than keep publishing versions.

Separately, this PR does not depend on the registry at all — the list entry
installs by git-clone. If the code is acceptable to you now, merging it would
make the node installable from the Manager at default security settings while
the registry side is sorted out.

Nothing outstanding from my side either way. 2.3.3 is the version to look at.

RebellionPictures:
2.3.4 is published, and it carries no findings.

Two things worth reporting.

First, the reason was findable all along. The versions endpoint takes an
undocumented query parameter:

  /nodes/{id}/versions?include_status_reason=true

Without it status_reason is omitted and a flagged version looks unexplained.
With it, the full scanner output comes back — rule, matched pattern, file, line
number, severity, and for banned versions your own written reasoning. That is
worth documenting somewhere publishers will find it. Four of us have open
issues on registry-backend asking what our findings were, and the answer was
one query parameter away the whole time.

Second, what it said. 2.3.2 and 2.3.3 had exactly one finding each:

  valyda_protokoll/pdf.py:85
      return bool(os.environ.get("VALYDA_ZELLPRUEFUNG"))

  rule      python_environment_manipulation
  pattern   $env_read2, matching the literal text "os.environ.get("
  severity  info

A debug switch. When set, the PDF builder writes a cell-geometry file beside
each document so a layout test can verify no text overflows its cell. It was
set only by the test suite, and the test suite is not in the published package
— so in anything a user installs, that read could never return true.

2.3.4 replaces it with a module-level flag. There is now no os.environ call
anywhere in the package, and status_reason on 2.3.4 is empty.

2.3.4 also fixes a real bug unrelated to any of this: projektkennung() strips
leading underscores, so the folder "_ohne_projekt" resolved to "ohne_projekt"
— an empty folder created beside the real one, and the Project node reported
"no records found". Folder names now go through one validated helper that
takes them as they lie on disk, with the same containment check.

I also read your reasoning on the two bans, which had not been posted here:

  2.3.0  ARBITRARY_FILE_LAUNCH via POST /valyda/oeffnen
  2.3.1  unsafe deserialization, attacker-reachable via node widget

Both are fair, both trace to the same unconfined path handling, and both were
fixed in 2.3.2. I would not have argued with either had I been able to read
them at the time.

So: path confinement in 2.3.2, English interface in 2.3.3, no findings at all
in 2.3.4. 2.3.4 is the version to look at.

ltdrdata:
Thanks for the fixes! I re-checked the repo — the write paths are now confined to the output/valyda base with resolved-path containment, and the UI is in English. I've registered VALYDA Protocol directly in `custom-node-list.json` (the list tail had drifted so the PR couldn't merge cleanly). Closing this so the entry isn't added twice; it's now available in ComfyUI-Manager.

PR #3151, opened 2026-08-11
Add MiniMax H3 Latent

JerryZRic:
Adds ComfyUI MiniMax H3 Latent to custom-node-list.json.\n\nRepository: https://github.com/JerryZRic/comfyui-minimax-h3-latent\n\nThis extension provides Save/Load nodes for MiniMax H3 AV NestedTensor latents, avoiding the built-in Save Latent limitation for H3 video/audio latent pairs.

### DIFF
--- a/custom-node-list.json
+++ b/custom-node-list.json
@@ -61018,6 +61018,17 @@
             ],
             "install_type": "git-clone",
             "description": "A ComfyUI node for image layering, blending, compositing, and batch processing."
+        },
+        {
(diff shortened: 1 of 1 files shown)


ltdrdata:
Thanks — this is now registered in custom-node-list.json (and node_db/new). I added the entry directly because the list tail had drifted from your branch and the PR could no longer merge cleanly. Closing this PR so the entry doesn't get added a second time; no further action needed on your side.

PR #3162, opened 2026-08-14
Add ComfyUI-SplatKit

mickmumpitz:
Adds ComfyUI-SplatKit to custom-node-list.json.
**What it does:** Builds 3D Gaussian Splat training datasets from a single 360° panorama, inside ComfyUI.
**Registry:** Already published to the Comfy Registry under https://registry.comfy.org/publishers/mickmumpitz/nodes/comfyui-splatkit

### DIFF
--- a/custom-node-list.json
+++ b/custom-node-list.json
@@ -61029,6 +61029,17 @@
             ],
             "install_type": "git-clone",
             "description": "OpenAI API based prompt rewriting agents for MiniMax H3 T2VA/I2VA/L2VA/FL2VA and Ref2VA workflows, using the official H3 skill files."
+        },
+        {
+            "author": "mickmumpitz",
+            "title": "ComfyUI-SplatKit",
+            "id": "splatkit",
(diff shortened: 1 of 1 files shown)


ltdrdata:
Following up after merge — a fuller review found a security issue, so I've moved this entry to the dev channel (out of default) until it's fixed, and will restore it once verified.

The Dataset Project node deletes a folder named by its `dataset_name` field with no path check: with `reset` on, `make()` runs `shutil.rmtree` on `os.path.join(output_dir, dataset_name)` (`nodes/common.py:169-173`), which drops the base on an absolute name and doesn't stop `..`. A workflow can be queued with no access control, so a remote caller can run this node with `reset` and a name that resolves outside the output folder, deleting a directory elsewhere on disk.

Please confine `dataset_name` in `_p2s_output_base`/`_output_base_nomake` with `realpath()` + `commonpath([resolved, output_dir]) == output_dir`, rejecting absolute/traversing names (covers `nodes/hires_dataset.py` too). Happy to re-verify and move it back once that's in.

PR #3168, opened 2026-08-14
Update custom-node-list.json for Enhanced Prompt Editor

pixelpainter:
Adds Enhanced Prompt Editor to the custom node list.

Repository: https://github.com/pixelpainter/comfyui-AI-prompt-editor
Comfy Registry: https://registry.comfy.org/nodes/comfyui-ai-prompt-editor

A prompt-writing node for ComfyUI driven by local Ollama models - prompt enhance, variations, natural-language instruct editing, style targeting, image/video to prompt captioning, a browser for Civitai and Genur.art prompts and workflows, and wireless prompt injection into any text widget.

GPL-3.0. Single entry added, no other changes.

### DIFF
--- a/custom-node-list.json
+++ b/custom-node-list.json
@@ -61029,6 +61029,17 @@
             ],
             "install_type": "git-clone",
             "description": "OpenAI API based prompt rewriting agents for MiniMax H3 T2VA/I2VA/L2VA/FL2VA and Ref2VA workflows, using the official H3 skill files."
+        },
+        {
+            "author": "pixelpainter",
+            "title": "Enhanced Prompt Editor",
+            "id": "comfyui-ai-prompt-editor",
+            "reference": "https://github.com/pixelpainter/comfyui-AI-prompt-editor",
+            "files": [
+                "https://github.com/pixelpainter/comfyui-AI-prompt-editor"
+            ],
+            "install_type": "git-clone",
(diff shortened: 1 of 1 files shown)


ltdrdata:
The `ollama/generate` route sends a request to a client-supplied URL and returns the response; its only gate compares Origin to Host, which does not stop an off-origin caller.
- api.py:4453 (fetch), api.py:1809 (gate)

Please restrict the fetch to a host allow-list. I'll re-evaluate once the destination is constrained.

pixelpainter:
The destination is an allow-list now.

Loopback is allowed with no configuration, and that's the only thing allowed by default. Any other host has to be named explicitly as `host:port`, either in `EPE_OLLAMA_ALLOWED_HOSTS` or in `ollamaAllowedHosts` in `<ComfyUI user dir>/epe/config.json`.

Nothing over HTTP can add to it — no route writes that file or the cache it's read into.

The existing address checks still run first, so the allow-list only narrows: listing a public host doesn't get a public fetch, and the link-local block on metadata remains. A malformed or unreadable config leaves loopback-only rather than failing open.

`api.py:4453` is unchanged apart from that — it was already a fixed `/api/generate` path against a resolved, redirect-disabled host rather than an arbitrary URL, so the change is entirely in `_epe_check_ollama_url`.

Regarding `api.py:1809` — the Origin/Host compare only stopped browsers, and a local process walks past it. That's why the destination gate doesn't rely on it.

pixelpainter:
This is in 1.0.21, just published.

ltdrdata:
Thanks for the contribution! I've registered `Enhanced Prompt Editor` directly in `custom-node-list.json` — the list tail had drifted from your branch's base, so the PR couldn't be merged cleanly. Closing this so the entry isn't added twice; it's now available in ComfyUI-Manager.

PR #3178, opened 2026-08-17
Add ComfyUI-FeiHou-Easy-H3

FX-FeiHou:
## Summary

- Adds the standard edition only; the RunningHub (RH) edition is intentionally not included.
- Registry ID: comfyui-feihou-easy-h3
- Repository: https://github.com/FX-FeiHou/ComfyUI-FeiHou-Easy-H3

## Validation

- Parsed custom-node-list.json successfully after adding the entry.

### DIFF
--- a/custom-node-list.json
+++ b/custom-node-list.json
@@ -1,6 +1,17 @@
 {
     "custom_nodes": [
         {
+            "author": "FX-FeiHou",
+            "title": "ComfyUI-FeiHou-Easy-H3",
+            "id": "comfyui-feihou-easy-h3",
+            "reference": "https://github.com/FX-FeiHou/ComfyUI-FeiHou-Easy-H3",
+            "files": [
+                "https://github.com/FX-FeiHou/ComfyUI-FeiHou-Easy-H3"
+            ],
+            "install_type": "git-clone",
+            "description": "MiniMax H3 generation nodes for ComfyUI with an embedded 9-image, 3-video, and 3-audio reference gallery, LoRA stack support, optional second-pass sampling, and prompt optimization."
+        },
+        {
             "author": "Carasibana",
             "title": "ComfyUI-SimpleFloatSlider",
             "reference": "https://github.com/Carasibana/ComfyUI-SimplayboyleFloatSlider",


ltdrdata:
An open settings route stores an API base URL with no host restriction, and a separate open route then fetches from that stored URL — so a remote caller can make the server request an arbitrary host.
- nodes.py:1639-1646 (settings write), nodes.py:1648-1664, nodes.py:997-999 (fetch)

Please restrict the fetch to an explicit host allow-list and make the settings route local-only. I'll re-evaluate once the fetch destination is constrained.

Additionally, please write the node UI strings in English (currently two node display names (RemixLoader, DurationCrop) and several execution-error messages are in Chinese).

FX-FeiHou:
Addressed the requested security and localization changes in the linked node repository (now released as v1.4.2):

- Custom API endpoints are validated against an explicit, local-only host allow-list. External endpoints require HTTPS; Ollama is loopback-only.
- Both model discovery and prompt requests use the validator and do not follow redirects to another host.
- The settings, model-discovery, prompt-optimization, and LoRA-list routes reject non-loopback requests.
- The allow-list is documented in the README and the API Settings UI.
- Default node labels and backend errors now use English; native ComfyUI locales provide English and Chinese definitions, including Remix Loader and Digital Human/MV Duration Crop.

Could you please re-evaluate the PR? Thank you.

FX-FeiHou:
Addressed the requested security and localization changes in the linked node repAddressed the requested security and localization changes in the linked node repository (now released as v1.4.2):

- Custom API endpoints are validated against an explicit, local-only host allow-list. External endpoints require HTTPS; Ollama is loopback-only.
- Both model discovery and prompt requests use the validator and do not follow redirects to another host.
- The settings, model-discovery, prompt-optimization, and LoRA-list routes reject non-loopback requests.
- The allow-list is documented in the README and the API Settings UI.
- Default node labels and backend errors now use English; native ComfyUI locales provide English and Chinese definitions, including Remix Loader and Digital Human/MV Duration Crop.

Could you please re-evaluate the PR? Thank you.

ltdrdata:
Thanks for the quick turnaround on v1.4.2 — this is a great update. I re-cloned at HEAD and re-ran the check, and the security side looks good now. Confirming what landed: custom API endpoints are validated against your local-only host allow-list, external endpoints require HTTPS, Ollama is loopback-only, and model discovery and prompt requests go through the validator and don't follow redirects. I also see the settings, model-discovery, prompt-optimization, and LoRA-list routes now reject requests that don't come from the local machine. The outbound hosts are pinned and the write paths are confined — nothing arbitrary left. Nice work.

The only thing still open is the localization. I know you switched the default node labels and backend errors to English and added the native ComfyUI locale definitions — thank you for that. But the re-scan still sees the frontend UI strings coming through primarily in Chinese, so it looks like some of the interface labels haven't fully moved over yet. Could you send the remaining UI strings in English, along with the locale link you used, so I can confirm the coverage on my side?

PR #3180, opened 2026-08-20
Add ComfyUI Prompt Warehouse

D33MO:
## Summary

Adds [ComfyUI Prompt Warehouse](https://github.com/D33MO/ComfyUI-Prompt-Warehouse) to the ComfyUI-Manager custom node list.

The node pack provides:
- A persistent prompt warehouse with grouping and random selection
- Single-line and multiline prompt utility nodes
- A workflow-persisted multi-LoRA loader
- A Save Image variant with safe, confirmed deletion of its output files
- English and Simplified Chinese localization

## Validation

- `custom-node-list.json` parses successfully as JSON
- Only `custom-node-list.json` is modified

### DIFF
--- a/custom-node-list.json
+++ b/custom-node-list.json
@@ -1,6 +1,16 @@
 {
     "custom_nodes": [
         {
+            "author": "D33MO",
+            "title": "ComfyUI Prompt Warehouse",
+            "reference": "https://github.com/D33MO/ComfyUI-Prompt-Warehouse",
+            "files": [
+                "https://github.com/D33MO/ComfyUI-Prompt-Warehouse"
+            ],
+            "install_type": "git-clone",
+            "description": "Prompt warehouse and workflow utility nodes for ComfyUI, including reusable prompts, multiline prompt editing, a persistent multi-LoRA loader, and deletable image output."
+        },
+        {
             "author": "Carasibana",
             "title": "ComfyUI-SimpleFloatSlider",
             "reference": "https://github.com/Carasibana/ComfyUI-SimplayboyleFloatSlider",


ltdrdata:
This pack registers network-reachable routes with no access control — one deletes output image files, another overwrites stored prompt files.
- routes.py:53 (delete), prompt_store.py:62 (write)

Please make these routes local-only, or remove them. I'll re-evaluate once they are no longer network-reachable.

Additionally, please write the node UI strings in English (currently a dropdown default value and the API/validation error strings are in Chinese).

ltdrdata:
Thanks for the update — the earlier issue is resolved. One remaining item before this can be registered: the UI strings are in Chinese — an i18n locale table (`web/i18n.js`) that auto-selects on the browser's language, plus hardcoded Chinese combo options (`nodes.py:43`, `web/prompt_warehouse.js:187`), a Chinese widget default (`prompt_store.py:90`), and status text (`routes.py:29`/`92`). Please write the UI strings in English (and ship any translation as a per-user opt-in rather than auto-binding on the browser language). Locale feature: https://github.com/Comfy-Org/ComfyUI/pull/6558

I'll re-evaluate once this is addressed.

D33MO:
Hi @ltdrdata, thanks for taking another look. All UI strings are now English by default, and Chinese is an explicit per-user opt-in controlled by ComfyUI's own locale setting (`Comfy.Locale`). The browser language is not used for locale selection.

The changes are now on master (v0.4.1):

* `web/i18n.js` — English is the default table; `zh` is activated only when the user has selected Chinese in ComfyUI. `document.documentElement.lang` is no longer used as the first-choice source and is now only a last-resort fallback.
* `nodes.py` — English placeholders and `label_on/off`; the Chinese `"全部"` value is retained only as a hidden legacy value so existing workflows continue to load. `locales/en/nodeDefs.json` maps it to `"All"`.
* `prompt_store.py` — English widget defaults and validation errors.
* `routes.py` — English status/error messages, and the write/delete routes are now loopback-only (403 otherwise).
* `locales/en/nodeDefs.json`, `locales/zh/nodeDefs.json` — shipped through ComfyUI's locale mechanism from ComfyUI#6558.

Repo: https://github.com/D33MO/ComfyUI-Prompt-Warehouse (master @ 946cce1, v0.4.1)

Could you re-evaluate when you have a moment?

ltdrdata:
Thanks for the quick turnaround, and for the detail on what changed. I re-cloned at HEAD and re-checked. The UI-language side is sorted now — English is the default and the Chinese strings only show when a user opts in through Comfy.Locale, which is exactly the right approach. Nice work.

One thing still open on the delete side. The POST /prompt-warehouse/delete-output-images route at routes.py:86 removes output files named by the request (target.unlink), and its only guard is a same-machine check. That check isn't enough on its own: any web page the user has open in their browser makes its requests from the user's own machine, so it sails right past a loopback test. The route stays an open route with no access control, reachable by any page the user happens to be visiting.

Two fixes together close it: confine the target to the output folder (resolve the real path and check it stays inside a fixed base with realpath + commonpath, and reject absolute paths and "..") so it can only ever delete inside the intended directory; and make the delete genuinely user-initiated / local-only rather than something an open page can trigger — or drop the route and keep the delete inside node execution.

---

Correction — I got the delete side wrong: _safe_output_file already resolves each target and checks it stays inside the output directory (relative_to(output_root)) before unlinking, so that route is confined. Nothing further is needed — I'm registering the pack now.

ltdrdata:
Thanks again for the fixes! I have registered ComfyUI Prompt Warehouse directly in `custom-node-list.json` at the correct position in the list (new git-clone entries go at the end of the list, above the copy/unzip entries, rather than at the top). Closing this so the entry is not added twice; it is now available in ComfyUI-Manager.

PR #3181, opened 2026-08-20
Remove ComfyUI-SuperNodes from custom-node-list.json

sonnybox:
ComfyUI-SuperNodes is also published on the Comfy Registry as `comfyui-supernodes` (publisher: sonnybox). The manual entry here uses the id `supernodes`, so Manager sees two distinct packs that both target the `ComfyUI-SuperNodes` directory. I am having trouble with version management / registry issues so I am hoping removing it here would fix it.


sonnybox:
I found the source of the error PR no longer needed

PR #3183, opened 2026-08-21
Add ComfyUI-BatchPromptLoader to custom node list

ReSerendipity:
Add ComfyUI-BatchPromptLoader to the ComfyUI-Manager custom node list.

- Repository: https://github.com/ReSerendipity/ComfyUI-BatchPromptLoader
- Node: BatchPromptLoader (BatchPromptReaderWithClip)
- Description: ComfyUI custom node that loads and encodes prompts from TXT files with fixed/increment/decrement/random modes for batch image generation.
- No external dependencies.

### DIFF
--- a/custom-node-list.json
+++ b/custom-node-list.json
@@ -25120,6 +25120,16 @@
             "description": "A useful comfyui node named LoadArtistTag for selecting artist tags, including 1000+ single-artist tags and 300 mixed-artists tags."
         },
         {
+            "author": "ReSerendipity",
+            "title": "ComfyUI-BatchPromptLoader",
+            "reference": "https://github.com/ReSerendipity/ComfyUI-BatchPromptLoader",
+            "files": [
+                "https://github.com/ReSerendipity/ComfyUI-BatchPromptLoader"
+            ],
+            "install_type": "git-clone",
+            "description": "ComfyUI custom node that loads and encodes prompts from TXT files with fixed/increment/decrement/random modes for batch image generation."
+        },
+        {
             "author": "gmorks",
             "title": "ComfyUI-SendToDiscord",
             "reference": "https://github.com/gmorks/ComfyUI-SendToDiscord",


ltdrdata:
The `log_folder` widget value is used to build a directory path that is then created, written to, and deleted from with no confinement, so a path outside the intended directory can be reached — a graph can write and delete `.json` files anywhere the server can. The `folder_path` widget likewise opens files for reading from an unconfined path.
- batch_prompt_loader.py:204-216 (write/delete), batch_prompt_loader.py:139-312 (read)

Please confine these to a fixed base directory (resolve the real path and verify it stays within the base). I'll re-evaluate once the paths are contained.

Additionally, please write the node UI strings in English (currently the node display name and widget field labels are in Chinese).

ReSerendipity:
Thanks for the review — both points are addressed in `ReSerendipity/ComfyUI-BatchPromptLoader@98a61ad`.

## 1. Path traversal (`folder_path` / `log_folder`)

Added `_confine_to_base()`, which resolves the real path (following symlinks via `realpath`, normalized with `normcase` for Windows) and requires the result to stay inside the ComfyUI root. It is applied to `folder_path` and `log_folder`, and also to `output_folder` since it feeds an existence probe.

- Absolute paths outside the root and relative paths escaping via `..` are rejected with an explicit error.
- Default values (`input/batch_prompts`, `user/default/batch_prompt_logs`, `output/`) resolve inside the root and are unaffected.
- Log clearing no longer deletes every `.json` in the folder — it only removes files that parse as a dict containing the `prompt_file` key, i.e. entries previously written by this node.

Verified with a traversal test: `../../../../Windows/System32`, `../../../etc`, `C:/Windows`, `C:/Users`, and mixed forms such as `input/../../../../Windows` are all blocked, while valid relative and in-root absolute paths resolve normally (9/9 cases pass).

**Behavior note:** prompt folders and log folders must now live under the ComfyUI directory. If you would rather have the confinement scoped to `input/` and `user/` specifically instead of the whole ComfyUI root, I'm happy to narrow it.

## 2. UI language

- Display name: `批量提示词加载器 (Batch Prompt Loader)` -> `Batch Prompt Loader`
- All widget labels translated: `Folder Path`, `Current Index`, `Recursive Scan`, `Reverse Order`, `Filename Pattern`, `Output Folder`, `Enable Logging`, `Log Folder`, `Clear Log On Session Start`, `Skip Existing Images`
- Console and exception messages translated too, so no user-facing string remains in Chinese.
- i18n via the locale feature is noted as a follow-up; shipping English-first keeps this change easy to review.

Ready for re-evaluation whenever you have a moment.

ReSerendipity:
Follow-up: the README is now in sync with this change (fd69658). The wording that allowed arbitrary absolute paths, and the out-of-tree `C:\prompts\SFW` example, were replaced with in-root paths. The path-constraint requirement is now documented under Installation.

ReSerendipity:
Thanks for the review — both points are now addressed further.

**Narrowed path scope (per your suggestion):** the allowed range is no longer the whole ComfyUI root. Each user-supplied path is now confined to its own dedicated base directory:

- `folder_path` → `<ComfyUI root>/input`
- `log_folder` → `<ComfyUI root>/user`
- `output_folder` → `<ComfyUI root>/output`

Implementation details:

- `_confine_to_base(base_dir, user_path, label)` resolves the candidate with `realpath` + `normcase` and rejects anything that doesn't land inside the corresponding base (absolute paths outside it, `..` escapes, and cross-sibling jumps such as `input/../../output` are all blocked).
- Legacy values written against the ComfyUI root (e.g. `input/batch_prompts`, `user/default/batch_prompt_logs`, `output/`) keep working via a prefix-stripping helper, so existing workflows don't break.
- Widget defaults were adjusted accordingly (`folder_path` default is now `batch_prompts`), and the three path widgets got tooltips stating their base directory.
- README documents the three-base constraint (v2.1.2).

Verified with a unit test suite (21 cases covering the three bases, legacy prefixes, and escape attempts — all passing) and an end-to-end smoke test on a fake ComfyUI tree (valid reads succeed, escape paths are rejected before any file access). Changes are on `ReSerendipity/ComfyUI-BatchPromptLoader@master` (commit `c9c9c76`).

**UI language:** already switched to English (labels + README), as noted in my previous comment.

PR #3189, opened 2026-08-22
Add Model Fetcher (comfyui-modelfetcher)

KoLt-Real:
Adds **Model Fetcher**, already published on the Comfy Registry: https://registry.comfy.org/nodes/comfyui-modelfetcher

- Repository: https://github.com/KoLt-Real/ComfyUI-ModelFetcher
- Registers no nodes: a top-bar button that reads the open workflow's Markdown notes, downloads the models they list into the right `models/` subfolder (extra_model_paths aware, resume, HuggingFace token kept to HF hosts only), flags likely duplicates, and relinks loader nodes to copies already on disk.
- No dependencies beyond what ComfyUI ships (`requests`, `aiohttp`); no build step; `comfy node validate` passes clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/custom-node-list.json
+++ b/custom-node-list.json
@@ -61154,6 +61154,17 @@
             ],
             "install_type": "git-clone",
             "description": "ComfyUI nodes for Krea 2 AnyPaint arbitrary-mask inpainting and outpainting"
+        },
+        {
+            "author": "KoLt-Real",
+            "title": "Model Fetcher",
+            "id": "comfyui-modelfetcher",
+            "reference": "https://github.com/KoLt-Real/ComfyUI-ModelFetcher",
+            "files": [
+                "https://github.com/KoLt-Real/ComfyUI-ModelFetcher"
+            ],
+            "install_type": "git-clone",
+            "description": "Top-bar button that reads a workflow's Markdown notes, downloads the models it needs into the right folder, and relinks loader nodes to copies you already own."
         }
     ]
 }


ltdrdata:
This pack exposes network-reachable routes with no access control: one writes files from a caller-supplied job, and a token route writes/clears a credentials file. A `url` is also fetched with no host allow-list.
- downloader.py:204 (fetch), downloader.py:242,267 (write), hf_token.py:86-94,114 (token write/delete)

Please make these routes local-only, restrict the fetch to a host allow-list, and confine the write path. I'll re-evaluate once constrained.

KoLt-Real:
All three points are addressed in v1.2.0 (main, commit 7ed72fa; published to the Registry), with a hardening follow-up in 1.2.1.

**Local-only routes.** Every `/cf_mf/*` handler is wrapped in `access.local_only` (fetcher/access.py): only loopback peers are served, read from the socket peername (`request.remote`), never from a header. Anyone else gets a 403. The only opt-out is an environment variable on the server, `CF_MF_ALLOW_REMOTE=1`, for people who drive ComfyUI from another machine.

**Host allow-list on the fetch.** fetcher/urlpolicy.py — `check_url` requires http(s), no credentials in the URL, and a host on the list (`huggingface.co`, `hf.co`, `civitai.com`, `github.com` and the CDNs they redirect to; the `HF_ENDPOINT` host; operator entries via `CF_MF_ALLOWED_HOSTS`). It is enforced in `/cf_mf/download`, in the size probe before any request, and again in the worker. `open_url` is now the only code that follows redirects, one hop at a time, re-checking every hop; `Authorization` is dropped on any origin change and never re-added, so the token never leaves HuggingFace.

**Confined writes.** The route already resolved the destination server-side from the category (`resolve_category` → `dest_dirs` → `_safe_join`); the worker now re-checks the final path itself with `scanner.is_allowed_dest` (every folder ComfyUI registered for models, `output/` excluded) before its first `os.makedirs`, and refuses anything else whatever it is handed. The token route only accepts a single printable line, validated against HF before it is written to the fixed path under ComfyUI's `user` folder.

Tests: `tests/test_routes.py` (the gate on all eight routes, forged headers, the env opt-out, host refusals), `tests/test_security.py` (allow-list incl. look-alikes and credentials, destination check) and `tests/test_downloader_resume.py` (redirects against a local server: followed on the same origin, refused off the list after a single request, destination outside the model folders refused before any I/O). The README has a Security section documenting the rules and the two environment variables.

ltdrdata:
Thanks for the contribution! I've registered `Model Fetcher` directly in `custom-node-list.json` — the list tail had drifted from your branch's base, so the PR couldn't be merged cleanly. Closing this so the entry isn't added twice; it's now available in ComfyUI-Manager.

PR #3192, opened 2026-08-22
Update snapshot.js Close Button

admdev24:
### Component-Specific Style Refinement of the ComfyUI Manager Dashboard

**Maintain the same style as the main UI**

**1. Snapshot Manager Interface Harmonization:**

- > **Original Style:** The top-right close icon (×) inside the Snapshot Manager dialog panel suffered from a layout styling exclusion bug. On idle, it was enclosed by a prominent, unstyled square framework box outline that broken the visual continuity of the interface. On cursor hover, the component remained completely transparent and lacked tactile response or alignment properties.

- **Updated Style:** Fully stripped away all inherited container borders, default box outlines, and edge tracking constraints to establish a transparent, boundary-free profile on idle. The button layout logic was corrected locally at the file entrypoint by embedding a dedicated, non-blocking <style> hook right at the top of snapshot.js to ensure the layout parameters register instantaneously when the popup script initializes.

**2. Localized Spatial Alignment Calibration**

- > **Original Style:** Standard browser formatting rules applied a rigid geometric center calculation to the font glyph character X. This technical alignment generated an optical illusion, making the close cross icon appear pushed out of bounds and visually unbalanced to the right of its relative interactive frame.

- **Updated Style:** Isolated the internal text character metrics from the background container layout layers to enable precision tuning across asynchronous framework modules. The character font bounds were scaled and center-locked cleanly, while an independent, floating pseudo-element (::before) layer generates an elegant, light-grey circular backplate (rgba(255, 255, 255, 0.15)) exclusively during interactive mouse hover states.

- **Precision Tuning:** Calibrated the horizontal transformation mapping specifically for the sub-manager panel layers to an eye-centered ratio lock (transform: translate(-55%, -50%)). This manual offset adjustment neutralizes the font rendering illusion, ensuring a visually balanced, pixel-perfect close button configuration across all sub-manager layout dialogs.

Before:

After (Hovering):

### DIFF
--- a/js/snapshot.js
+++ b/js/snapshot.js
@@ -4,6 +4,76 @@ import { ComfyDialog, $el } from "../../scripts/ui.js";
 import { manager_instance, rebootAPI, show_message, handle403Response, loadCss } from  "./common.js";
 import { buildGuiFrame } from "./comfyui-gui-builder.js";
 
+// =========================================================================
+// PERMANENT FIX FOR SNAPSHOT MANAGER CLOSE BUTTON (LOCAL SOURCE HOOK)
+// =========================================================================
+const snapshotCustomStyle = document.createElement('style');
+snapshotCustomStyle.innerHTML = `
+    /* 1. Clear out the unstyled bounding box and borders on idle */
+    [id*="snapshot"] button[class*="close"],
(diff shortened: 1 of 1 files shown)


admdev24:
Will edit the snapshot.css instead.

PR #3203, opened 2026-08-24
Add ComfyUI Video Split Nodes by mickeylan

mickeylan:
A comprehensive video processing node library for AI manga production. Includes: video segmenting/merging, audio processing (extract/mix/fade-in-out), text/subtitle overlay with Chinese font support, filters/color grading, transitions, effects (background removal/replacement), AI assistant (auto subtitle with Whisper, auto TTS with Edge-TTS), Wan22/LTX long-video chunked sampling for 720p high-resolution on 12GB VRAM, batch render queue, and 60+ nodes in total.

### DIFF
--- a/custom-node-list.json
+++ b/custom-node-list.json
@@ -61163,7 +61163,7 @@
                 "https://github.com/mickeylan/comfyui-video-split"
             ],
             "install_type": "git-clone",
-            "description": "A comprehensive video processing node library for AI manga production. Includes: video segmenting/merging, audio processing (extract/mix/fade-in-out), text/subtitle overlay with Chinese font support, filters/color grading, transitions, effects (background removal/replacement), AI assistant (auto subtitle with Whisper, auto TTS with Edge-TTS), Wan22/LTX long-video chunked sampling for 720p high-resolution on 12GB VRAM, batch render queue, and 60+ nodes in total."
+            "description": "A comprehensive video processing node library for AI manga production. Includes: video segmenting/merging, audio processing (extract/mix/fade-in-out), text/subtitle overlay with Chinese font support, filters/color grading, transitions, effects (background removal/replacement), AI assistant (auto subtitle with Whisper, auto TTS with Edge-TTS), Wan22/LTX long-video chunked sampling for high-resolution video generation, batch processing, and 60+ nodes in total."
         }
     ]
 }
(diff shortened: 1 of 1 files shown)


ltdrdata:
1. The translator nodes send data to a URL taken from a widget (`ollama_url` / `deeplx_url`) with no host restriction, so the request can be pointed at an arbitrary host.
   - translation_nodes.py:495-518
2. The audio/video path widget is read from an unconfined absolute path and its content is returned via the transcript output, allowing arbitrary files to be read.
   - translation_nodes.py:831-838, translation_nodes.py:955-964

Please restrict the outbound host to an allow-list and confine the read path to a fixed base directory. I'll re-evaluate once both are constrained.

Additionally, please write the node UI strings in English (currently the tooltips, DESCRIPTION fields, dropdown options, and JS UI are in Chinese).

ltdrdata:
Thanks for the update, and for turning this around so quickly!

I re-cloned at your latest commit and re-ran the check. Good news on most of it: the other file operations and the ffmpeg calls are properly contained now — the output/save side lands inside the intended folder, so that part is sorted.

There are two spots left, both on the read side, where a path is still taken in without containment:

1. translation_nodes.py:832 — AudioReplacer.replace resolves video_path with os.path.abspath and accepts absolute paths, so a supplied path can read files outside the ComfyUI input folder.
2. translation_nodes.py:955 — VideoTranslatePipeline.process resolves video_path the same uncontained way and then transcribes it, so the contents of an arbitrary readable file can come back out through the transcript/srt outputs.

So the save side is fixed; it's just these two reads that still resolve the path freely. The remedy is the same containment you already applied elsewhere: resolve each path against a fixed base directory (the ComfyUI input folder) using realpath plus a commonpath check, and reject absolute paths and any "../" segments. Fixing it at the path-resolution step covers both nodes.

---

Update — one more item from a fuller re-check: FinalVideoSave (ltxv_stream_save_nodes.py:395-439) trusts session_dir and the segment paths from its VIDEO_SEGMENT_COLLECTION input without containment, then writes segments.txt there, runs ffmpeg concat with -safe 0 over the listed paths, and calls shutil.rmtree(session_dir). ComfyUI does not type-check a literal value for a custom input type, so that input can be supplied directly in a prompt. Please confine session_dir (and every listed segment path) to the temp directory your own nodes create, with realpath + commonpath, before writing, reading or deleting.

PR #3205, opened 2026-08-25
Add ComfyUI Free Model and Node Cache to custom-node-list.json

VRAM-Hoarder:
A ComfyUI custom node that replicates the function of the 'Free model and node cache' toolbar button - usable inline in a workflow, so you can force a VRAM/RAM cleanup automatically after every generation instead of clicking a button manually.


ltdrdata:
Due to merge conflicts, I've added the entry for `ComfyUI Free Model and Node Cache` to `custom-node-list.json` directly, using the title and description from this PR.

PR #3221, opened 2026-08-29
Add Unbake

syugoji:
Unbake rebuilds the ComfyUI workflow behind an image — a Civitai link, a PNG with
metadata, or ComfyUI's own output — reads the generation parameters, sorts the missing
models by how you would obtain them, and then runs a parameter sweep that refuses to
start if anything other than the axis you declared would move.

It adds one canvas node (`UnbakeRecipeSource`) plus a sidebar panel. GPL-3.0, no pip
dependencies. It is already on the registry as `comfyui-unbake`; this entry is so the
list carries the GitHub reference.

Repository: https://github.com/syugoji/ComfyUI-Unbake

### DIFF
--- a/custom-node-list.json
+++ b/custom-node-list.json
@@ -1,6 +1,16 @@
 {
     "custom_nodes": [
         {
+            "author": "syugoji",
+            "title": "Unbake",
+            "reference": "https://github.com/syugoji/ComfyUI-Unbake",
+            "files": [
+                "https://github.com/syugoji/ComfyUI-Unbake"
+            ],
+            "install_type": "git-clone",
+            "description": "Drop an image and get the ComfyUI workflow back: a Civitai link, a PNG with metadata, or your own output. Unbake reads the PNG info — prompt, seed, sampler, steps, CFG, LoRAs — rebuilds the graph, and sorts the missing models by how you would obtain them. Then it runs a validated XY plot (parameter sweep) that throws if anything other than the axis you declared has moved, so the comparison you are looking at is actually a comparison."
+        },
+        {
             "author": "Carasibana",
             "title": "ComfyUI-SimpleFloatSlider",
             "reference": "https://github.com/Carasibana/ComfyUI-SimplayboyleFloatSlider",


ltdrdata:
The `model-delete` route deletes model files and is network-reachable, so any page the user visits can delete files.
- unbake/models.py:297

Please make this route local-only, or remove it. I'll re-evaluate once it is no longer network-reachable.

syugoji:
Thanks — fixed in syugoji/ComfyUI-Unbake@3dae74a, and I found two more of the same shape.

You named `model-delete`, but `record-delete` and `output-delete` were reachable the same way, and there was no origin or locality check anywhere in the package. Rather than patch the three, the check now lives in one place (`unbake/guard.py`) and every POST route carries it. A test reads the source and fails if a POST route is added without it.

One note on the fix: local-only alone does not stop what you described. A malicious page runs in the user's own browser, so its request arrives from 127.0.0.1 like any other. The guard therefore requires all three of:

- the peer to be loopback — stops other machines on the network, since ComfyUI binds 0.0.0.0 by default;
- `Origin`, when present, to match the host — stops the cross-site case you described;
- `Content-Type: application/json` — forces a CORS preflight. Without it a `text/plain` body is a "simple request", skips the preflight, and walks past the origin check.

A request with no `Origin` is still allowed, since browsers always send one on a cross-origin POST and the loopback condition covers command-line callers. `X-Forwarded-For` is ignored on purpose — the sender picks its value, so trusting it would undo the loopback check. Requests through a reverse proxy or tunnel are now refused, which I took to be the intent.

Tests cover both the wiring (every POST route carries the guard) and the behaviour (each condition refuses on its own). I mutated each one to confirm the tests fail without it.

The route is no longer network-reachable. Happy to make any of it stricter if you would prefer.

ltdrdata:
Thanks for the update — the earlier issue is resolved. One remaining item before this can be registered: the pack ships a 12-language i18n locale table (`web/i18n/index.js` + `web/i18n/locales/{ja,ko,zh,zh-TW,ru,ar,fa,tr,fr,es,pt-BR}.js`) that auto-selects on the browser's language, so on a non-English browser the UI renders in that language. For the default channel the UI needs to stay English — please ship the translations as a per-user opt-in rather than auto-binding on the browser language, or keep the UI English-only. For multilingual support, refer to the locale feature: https://github.com/Comfy-Org/ComfyUI/pull/6558

I'll re-evaluate once this is addressed.

syugoji:
Thanks for taking another look.

On the i18n point I think there may be a misunderstanding of what it binds to, so let me be precise rather than argue.

It does not read the browser language. There is no `navigator.language`, `navigator.languages` or `Accept-Language` anywhere in the package — the only input is ComfyUI's own setting:

```
web/unbake.js:2621
readHostLocale(app) -> app.ui.settings.getSettingValue('Comfy.Locale')
```

That is the locale feature you linked. The catalogue list is deliberately the same twelve entries `Comfy.Locale` offers, so the panel cannot offer a language the host cannot, and cannot lack one the host can. If the setting is unreadable the panel falls back to English (`web/i18n/index.js:64`, `DEFAULT_LOCALE = 'en'`), so English is the default in every case where the host has not said otherwise.

The per-user opt-in you are asking for is already there as well: a `language` setting, default `""` meaning "follow the host", with a selector in the panel that rebuilds the UI on change (`unbake/settings.py:272` and `web/unbake.js:1287`).

The reason it follows the host rather than offering an independent switch is that an independent switch lets a user reach a state where ComfyUI is in one language and this one panel is in another, which is worse than either being consistently wrong.

So the one thing I cannot decide for you: with the above, a user who has set ComfyUI itself to Japanese gets this panel in Japanese without opting in separately. If your requirement is that the default channel renders English until the user explicitly picks a language — regardless of what the host is set to — say so and I will make English the initial value and let only the `language` setting move it. That is a small change and I would rather match your policy than guess at it.

ltdrdata:
Thanks for the contribution! I've registered `Unbake` directly in `custom-node-list.json` — the list tail had drifted from your branch's base, so the PR couldn't be merged cleanly. Closing this so the entry isn't added twice; it's now available in ComfyUI-Manager.

PR #3239, opened 2026-09-03
Add ValiTools node pack

vangel76:
Adds the ValiTools node collection to `custom-node-list.json`.

The pack is published on the Comfy Registry but is not in this list yet, so it shows up without stars or last-update information in the Manager.

**Nodes**
- **VSmartPrompt** — Dynamic Prompts with a rich-text editor: combinations with weights, wildcards, reusable variables (rolled once, constant everywhere), switcher conditions, chaining inputs, and post-run highlighting of the branches that were actually selected
- **VFileRandom** — random image loader that never repeats until every image in the folder has been drawn
- **VRandomSelector** — lazy random passthrough of N same-type inputs; only the selected branch executes
- **VWaitForVRAM** — holds execution until the GPU has a minimum amount of free VRAM

Repository: https://github.com/vangel76/comfyui-ValiTools
Registry: https://registry.comfy.org/publishers/vangel/nodes/valitools
License: Apache-2.0

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/custom-node-list.json
+++ b/custom-node-list.json
@@ -1,6 +1,16 @@
 {
     "custom_nodes": [
         {
+            "author": "Valentin Klein",
+            "title": "ValiTools",
+            "reference": "https://github.com/vangel76/comfyui-ValiTools",
+            "files": [
+                "https://github.com/vangel76/comfyui-ValiTools"
+            ],
+            "install_type": "git-clone",
+            "description": "ValiTools node collection: VSmartPrompt (Dynamic Prompts with a rich-text editor, reusable variables, switcher conditions, chaining inputs and post-run selection highlighting), VFileRandom (no-repeat random image loader), VRandomSelector (lazy random passthrough of N same-type inputs) and VWaitForVRAM (hold execution until enough VRAM is free)."
+        },
+        {
             "author": "Carasibana",
             "title": "ComfyUI-SimpleFloatSlider",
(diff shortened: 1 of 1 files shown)


ltdrdata:
1. The wildcard routes resolve their containment root from a client-supplied `wildcard_dir` field in the request body instead of a fixed base directory, so the `is_relative_to()` check runs against an attacker-chosen root and confines nothing — these open routes with no access control can then read any `.txt` file and write attacker content to any `.txt` path (both also creating parent directories on the way).
   - nodes.py:1619-1645 (root resolution), nodes.py:1680-1708 (save_wildcard write), plus read_wildcard / get_wildcard_files / validate_wildcards
2. The VFileRandom `folder` widget and the VSmartPrompt `wildcard_directory` widget are read from an unconfined path, and VSmartPrompt exposes the read content through its output.

Please resolve every route path against a fixed base directory (`realpath` + `commonpath` against a constant wildcard root, rejecting absolute and `..` paths), ignore the client-supplied directory field, and confine the two widget reads the same way. I'll re-evaluate once the paths are contained.

---

Update — one more item from a fuller re-check: VFileRandom (vfile_random.py:134) reads image files from any directory given in its folder widget and returns them as IMAGE output. Please confine folder to a fixed base (e.g. the input directory) with realpath + commonpath, rejecting absolute paths and "..", as for the wildcard routes.

PR #3241, opened 2026-09-04
Add Autocomplete++ to custom node list

michikora:
`Author`: michikora
`Title`: Autocomplete++
`Description`: Tag autocomplete, prompt formatting, and dynamic prompt expansion extension for ComfyUI.
`Repository`: https://github.com/michikora/ComfyUI-Autocomplete-Plus-Plus
`Install type`: git-clone

### DIFF
--- a/custom-node-list.json
+++ b/custom-node-list.json
@@ -61469,6 +61469,16 @@
             ],
             "install_type": "unzip",
             "description": "This is a node to convert an image into a CMYK Halftone dot image."
+        },
+        {
+            "author": "michikora",
+            "title": "Autocomplete++",
+            "reference": "https://github.com/michikora/ComfyUI-Autocomplete-Plus-Plus",
+            "files": [
+                "https://github.com/michikora/ComfyUI-Autocomplete-Plus-Plus"
+            ],
+            "install_type": "git-clone",
+            "description": "Tag autocomplete, prompt formatting, and dynamic prompt expansion extension for ComfyUI."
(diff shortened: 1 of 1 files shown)


ltdrdata:
Thanks for the contribution! Before this can be registered, one issue needs addressing:

1. The `GET /autocomplete-plus-plus/tags/file/{filename}` and wildcard-content routes resolve a caller-supplied name into a file path with only a `replace('\\','/').strip('/')` cleanup and no base-directory containment, so a remote caller can read `.txt` files outside the intended wildcard/tag directories.
   - modules/api.py:224 (get_wildcard_content), and the related file/tag routes

Please resolve every route path against a fixed base directory (`realpath` + `commonpath` against the intended root, rejecting absolute and `..` paths) before opening the file.

I'll re-evaluate once the paths are contained.

michikora:
Hello, Thanks for taking the time to review!

I've addressed the issue by adding a helper function `resolve_contained_file` to enforce base-directory containment (`realpath` + `commonpath`, rejecting absolute paths and `..`). 
The following routes have been updated:
- `GET /autocomplete-plus-plus/tags/file/{filename}`
- `GET /autocomplete-plus-plus/wildcards/content`
- `GET /autocomplete-plus-plus/wildcards/all-data`
- `GET /autocomplete-plus-plus/models/thumbnail`

Please take another look when you have time to see if everything looks good. 

Thanks!

PR #3247, opened 2026-09-06
Add MSCH Nodes as one unified custom-node pack

mariobilly:
Register **MSCH Nodes** as one installable pack at https://github.com/mariobilly/msch-comfyui-nodes.

The repository contains all 35 existing MSCH node IDs and the theme/editor bridge in one package. Node implementations, full references, workflow examples and rendered showcases are organized by component. The root loader isolates component failures and detects known legacy installations to avoid duplicate node, route and frontend registrations. MIGRATION.md explains switching from the former separate packs without changing workflow node IDs.

Validation: the actual ComfyUI custom-node loader registered all 35 nodes under the unified package; all 15 components and six frontend entry points loaded; HTTP checks covered the entry points, editor, font and sequencer assets; Code Matrix and Typort smoke renders passed; six loader tests and all 108 existing A2V core tests passed. GitHub checks also validate syntax, metadata, workflow JSON and browser import paths.

This submission adds one Manager entry. The earlier proposed 15 separate entries have been replaced in this branch. Historical repositories link to the new pack for migration.

### DIFF
--- a/custom-node-list.json
+++ b/custom-node-list.json
@@ -1,6 +1,156 @@
 {
     "custom_nodes": [
         {
+            "author": "mariobilly",
+            "title": "MSCH A2V",
+            "reference": "https://github.com/mariobilly/msch-a2v",
+            "files": [
+                "https://github.com/mariobilly/msch-a2v"
+            ],
+            "install_type": "git-clone",
+            "description": "Beat-synced prompt sequencing, shot planning, MiniMax H3 sampling, assembly and pixel upscaling for music videos."
+        },
+        {
+            "author": "mariobilly",
+            "title": "MSCH Code Matrix",
+            "reference": "https://github.com/mariobilly/msch-code-matrix",
+            "files": [
+                "https://github.com/mariobilly/msch-code-matrix"
+            ],
+            "install_type": "git-clone",
+            "description": "Convert images and video frames into ASCII portraits, binary art and animated Matrix-style glyph patterns."
+        },
+        {
+            "author": "mariobilly",
+            "title": "MSCH Edge ASCII",
+            "reference": "https://github.com/mariobilly/msch-edge-ascii",
+            "files": [
+                "https://github.com/mariobilly/msch-edge-ascii"
+            ],
+            "install_type": "git-clone",
+            "description": "Contour-aligned ASCII strokes, halftone fills and animated marks for photos and native ComfyUI video."
+        },
+        {
+            "author": "mariobilly",
+            "title": "MSCH Lyric Sync",
+            "reference": "https://github.com/mariobilly/msch-lyric-sync",
+            "files": [
+                "https://github.com/mariobilly/msch-lyric-sync"
+            ],
+            "install_type": "git-clone",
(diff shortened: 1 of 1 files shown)


ltdrdata:
Thanks for the contribution! A few issues need addressing before this can be registered:

1. Several nodes take a full-path STRING widget and use it verbatim when absolute (only `is_file`/`is_dir` checks), so a shared workflow can read or write files anywhere on the host:
   - components/puppet_face/nodes.py:456 (write via cv2.VideoWriter), 402 (read via cv2.VideoCapture)
   - components/typort/nodes.py:55, components/slideshow/nodes.py:234 and 34, components/slideshow/advanced_nodes.py:21, components/a2v/mscha2v/nodes/_common.py:10
   Please resolve these against a fixed base directory (`realpath` + `commonpath`, rejecting absolute and `..`).
2. components/lyric_sync/aligner.py:40 takes a `whisperx_exe` widget and runs it as an executable if the path exists — an arbitrary-executable launch. Please pin this to a known binary / PATH lookup rather than an arbitrary caller-supplied path.
3. components/typort/audio_analysis.py:80 calls `torch.load(..., weights_only=False)` (an unsafe pickle deserializer) and is reachable from the open `POST /mariotyport/analyze-audio` route, so a crafted checkpoint executes code on load. Please load with `weights_only=True` (or a safetensors path).

mariobilly:
Thanks @ltdrdata. All three issues are addressed and pushed in [f59c2d4](https://github.com/mariobilly/msch-comfyui-nodes/commit/f59c2d42d478671935deccd9851ae30b167832a1).

1. **File paths:** all the listed PuppetFace, Typort, Slideshow and A2V entry points now use a shared resolver with `realpath` + `commonpath`. Reads are confined to ComfyUI/input and PuppetFace writes to ComfyUI/output. Absolute paths, Windows drive/UNC paths, `..` components and symlink escapes are rejected. Cache checks use the same validation, and Slideshow checks individual image entries as well as the directory. Related soundtrack/font inputs and the Typort audio route are also constrained. Tooltips and migration documentation explain the new input-relative paths.
2. **WhisperX:** executable selection now uses only the fixed `whisperx` command on the host PATH or the known installation fallback. The legacy `whisperx_exe` widget is ignored, including at the subprocess runner, so existing workflows cannot override the executable.
3. **Demucs:** checkpoint loading now uses `weights_only=True`, with only the fixed metadata types required by the official artifact allowlisted. The pinned SHA-256 is checked against the same in-memory bytes that are deserialized; there is no unsafe fallback. Tests confirm that a crafted pickle is rejected and that the official model still reconstructs successfully.

[GitHub Actions passed](https://github.com/mariobilly/msch-comfyui-nodes/actions/runs/34514669276), including all 24 loader/security tests on Linux (with the symlink cases and official checkpoint) and all 108 A2V tests. Locally, all 17 Typort and 28 Slideshow tests also passed. The actual ComfyUI loader registered all 35 nodes across 15 components; browser asset checks and a typography render passed.

Ready for re-evaluation. Thanks for identifying these issues.

PR #2705, opened 2026-03-18
add DemonAlone-miniPaint-ComfyUI

DemonAlone:

PR #3249, opened 2026-09-07
add DemonAlone-miniPaint-ComfyUI

DemonAlone:
## Security & compatibility update for miniPaint integration

This PR re-submits the integration after the previous one #2705  was closed due to security concerns in the verification scan.

All identified issues have been addressed:
- Ran `nodesafe` and `semgrep` scanners – no detections remain.
- Fixed all flagged vulnerabilities.
- Added an **iframe mode** alongside the existing "open in new window" mode – users can now choose their preferred workflow.
- Ensured full compatibility with **ComfyUI Node 2.0** (tested and verified).

### DIFF
--- a/custom-node-list.json
+++ b/custom-node-list.json
@@ -45025,6 +45025,16 @@
             "description": "Interactive JavaScript nodes: Load Video with real-time preview, Video Maker (image batch to MP4), plus upcoming queue management and dynamic UI tools. Enhances workflow flexibility."
         },
         {
+            "author": "DemonAlone",
+            "title": "DemonAlone-miniPaint-ComfyUI",
+            "reference": "https://github.com/DemonAlone/DemonAlone-miniPaint-ComfyUI",
+            "files": [
+                "https://github.com/DemonAlone/DemonAlone-miniPaint-ComfyUI"
+            ],
+            "install_type": "git-clone",
+            "description": "A lightweight, local, open-source image editor (miniPaint) seamlessly integrated into ComfyUI. Supports both iframe and new-window modes, full compatibility with Node 2.0."
+		},
+        {
             "author": "tabisheva",
             "title": "comfyui-segs-profile",
             "reference": "https://github.com/tabisheva/comfyui-segs-profile",


ltdrdata:
Thanks for the contribution! Two things need addressing before this can be registered:

1. `GET /mypaint/{file:.*}` (__init__.py:67) takes the wildcard path from the request and reads it with no base-directory containment — the `{file:.*}` pattern captures `..` segments — so a remote caller can read arbitrary files. Please resolve the served path against a fixed base directory (`realpath` + `commonpath`, rejecting absolute and `..`).
2. The bundled miniPaint editor (`editor_static/bundle.js`) ships i18n locale tables (Russian, Korean, Japanese, …) that auto-select on the browser language, so the served UI renders in a non-English language. For the default channel the UI needs to stay English — please serve an English-only build, or ship the translations as a per-user opt-in. For multilingual support, refer to the locale feature: https://github.com/Comfy-Org/ComfyUI/pull/6558

DemonAlone:
@ltdrdata , thank you for your review!

Both issues have been addressed.

1. Path traversal in GET /mypaint/{file:.*}
   The served path is now resolved with os.path.realpath and validated with
   os.path.commonpath against the fixed static base directory. Absolute paths
   and any path escaping the base are rejected with 403.

   Verified with:
     $ curl --path-as-is -i "http://127.0.0.1:8188/mypaint/../../__init__.py"
     HTTP/1.1 403 Forbidden
     $ curl -i "http://127.0.0.1:8188/mypaint/%2e%2e/%2e%2e/__init__.py"
     HTTP/1.1 403 Forbidden

   Normal requests (e.g. /mypaint/index.html) still return 200.

2. Non-English UI
   The editor is now pinned to English by default. Before bundle.js is loaded,
   we set localStorage.lang = "en" and override navigator.language /
   navigator.languages. The UI no longer auto-switches based on the browser
   locale. Other locales remain in the bundle but are only activated if the
   user explicitly changes the language in the editor (per-user opt-in).

   Everything should be fine now, since v2.3.5

PR #3250, opened 2026-09-07
Add ComfyUI-ColorID-Mask to extension-node-map

asamitsengupta:
Added a new entry for ComfyUI-ColorID-Mask with details.


ltdrdata:
Thanks! `extension-node-map.json` is generated automatically by our scanner from installed nodes rather than maintained through PRs, so this change can't be merged directly. To have the nodepack listed in ComfyUI-Manager, please open a PR adding an entry to `custom-node-list.json` instead.

PR #3254, opened 2026-09-08
Add QBB NodeOps to custom-node-list

rhinoflavored:
## Summary
- Add **QBB NodeOps** with its renamed repository URL.
- The previous submission (#3095) was closed due to two unrestricted server-side export paths. This version removes those paths: exports are now browser downloads with user-controlled save location.

## Node
- Repo: https://github.com/rhinoflavored/ComfyUI_QBB_NodeOps
- Comfy Registry publisher: 
hinoflavored
- Install type: git-clone

## Test plan
- [ ] Search Manager for QBB NodeOps or qbb-nodeops.
- [ ] Install via Manager, restart ComfyUI, and confirm the QBB NodeOps menu is available.


ltdrdata:
Thanks for the contribution! This pack exposes several open routes (no access control) that take dangerous actions from request input — these need addressing before it can be registered:

1. `POST /qbb_nodeops/ai/settings` lets a caller change the AI `base_url` while the stored API key is kept, so a subsequent `/ai/analyse|recommend|...` call sends `Authorization: Bearer <your stored key>` to the caller-chosen host — the operator's provider key is sent to an arbitrary host and the response is returned to the caller. Please restrict the AI base_url to an explicit allow-list and never send the stored key to a non-allow-listed host.
   - qbb_nodeops/ai_service.py:89
2. `POST /qbb_nodeops/network_settings` writes a caller-supplied proxy host/port and applies it as HTTP_PROXY/HTTPS_PROXY for the whole ComfyUI process and its subprocesses (persisted and re-applied at startup), so all outbound traffic — downloads, git, API calls — can be routed through a caller-chosen proxy. Please don't let a network route set a process-wide proxy; make this local-only / user-initiated.
   - qbb_nodeops/settings.py:105
3. `POST /qbb_nodeops/updates/pull` and `/updates/refresh` run `git fetch`/`git pull`/`git ls-remote` against installed plugins, and the fetch URL is rewritten with a `mirror_prefix` that is itself caller-settable via `/network_settings` — so a caller can point the fetch at an arbitrary host and have attacker-served history fast-forwarded into a plugin directory (code then runs on the next import). Please make update actions local-only / user-initiated, and constrain any mirror prefix to an allow-list.
   - qbb_nodeops/update_check.py:486 and 314
4. `POST /qbb_nodeops/migration/start` (with delete_source) calls `shutil.rmtree` on directories chosen through caller-registered instances, `/migration/scan` enumerates an arbitrary directory tree, and `POST /qbb_nodeops/instances` accepts an arbitrary path with no trusted-root containment (this feeds the delete/scan targets). Please confine these to a designated trusted root (`realpath` + `commonpath`) and make the destructive/migration actions local-only / user-initiated.
   - qbb_nodeops/migration.py:230 and 107, qbb_nodeops/instances.py:103
5. `GET /qbb_nodeops/environment` runs `pip freeze` and returns the full environment/dependency listing to any caller. Please make this local-only / user-initiated.
   - qbb_nodeops/env_export.py:11
6. The JS UI ships a Chinese locale table (`web/i18n.js`) that auto-selects on the browser language, plus a hardcoded Chinese `confirm()` dialog and a menu label (`web/nodeops.js`), and a Chinese error string returned to the UI (`qbb_nodeops/routes.py:166`). Please write the UI strings in English. For multilingual support, refer to the locale feature: https://github.com/Comfy-Org/ComfyUI/pull/6558

PR #3255, opened 2026-09-09
Add ComfyUI-Arisu-Nodes to custom-node-list

swqa7697:
Adds **ComfyUI-Arisu-Nodes** (`arisu_nodes`), a V3-API node pack of workflow-utility nodes.

- Repository: https://github.com/swqa7697/ComfyUI-Arisu-Nodes
- Registry: https://registry.comfy.org/nodes/arisu_nodes


ltdrdata:
Thanks for the contribution! A few issues need addressing before this can be registered:

1. `ArisuLoadImage` feeds its `path` STRING widget through a resolver that only does `expanduser`/`join(input_dir, path)`/`abspath` with no containment (src/arisu_nodes/common/nodes.py:413 → core.py:330), so a shared workflow can read any file on the host.
2. The open routes `GET /arisu/view` (routes.py:177) and `GET /arisu/browse` (routes.py:97) resolve the query `path` with only `isabs`/extension checks and no base-directory containment, so a remote caller can read arbitrary files and list arbitrary directories.
   Please resolve every path (widget and route) against a fixed base directory (`realpath` + `commonpath`, rejecting absolute and `..`).

I'll re-evaluate once the paths are contained.

swqa7697:
@ltdrdata Thanks for reviewing my project! The issues you pointed out are resolved in `v1.1.2`, and the regression tests covering these cases now pass. I’d appreciate another review when you have a chance. Thank you!

PR #3268, opened 2026-09-11
Add Antimeter YuE2 Local Runtime

AntiMatterComfy:
Adds the public AntiMatterComfy YuE2 integration to the Manager catalogue. The Manager installation clones only the custom-node package. YuE2's separately documented Windows WSL or native Linux installer performs the explicit local GPU check and installs the official runtime; it is not run automatically by Manager.\n\nRepository: https://github.com/AntiMatterComfy/antimeter-yue2-local-runtime\nRecommended release: https://github.com/AntiMatterComfy/antimeter-yue2-local-runtime/releases/tag/v1.1.2\n\nThe node has no bundled weights. Upstream YuE2 model weights remain under CC BY-NC 4.0.

### DIFF
--- a/custom-node-list.json
+++ b/custom-node-list.json
@@ -11,6 +11,18 @@
             "description": "ComfyUI custom nodes providing styled draggable float slider widgets"
         },
         {
+            "author": "AntiMatterComfy",
+            "title": "Antimeter YuE2 Local Runtime",
+            "id": "antimeter-yue2-local-runtime",
+            "nodename_pattern": "AntimeterYuE2",
+            "reference": "https://github.com/AntiMatterComfy/antimeter-yue2-local-runtime",
+            "files": [
+                "https://github.com/AntiMatterComfy/antimeter-yue2-local-runtime"
+            ],
+            "install_type": "git-clone",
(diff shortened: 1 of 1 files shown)


ltdrdata:
Thanks for the contribution!

I took a look at antimeter-yue2-local-runtime and found one thing I'd like fixed before it can go in:

1. The music-runner node launches whatever program the `runner_path` widget points at. On the Linux path it calls `subprocess.run([str(runner), ...])` (__init__.py:147), and on the Windows/WSL path it builds `sh -lc` with the same widget as the exec target (__init__.py:142). The value comes from a node widget, which any workflow can set, and the only check is that the path exists — so any workflow could make ComfyUI run an arbitrary program on the host. Please don't let the request choose the executable: pin it to a fixed runner, or check the requested path against an explicit allow-list of permitted runners, before invoking it. (The shlex-quoting on the WSL path guards the arguments, but the executable itself still needs to be constrained.)

For what it's worth, the nvidia-smi GPU preflight (__init__.py:81) is fine — it uses a fixed executable with fixed arguments and a validated distro name, so no change needed there.

PR #3266, opened 2026-09-11
Add ComfyUI Prompt Workbench and LoRA Trigger Helper

yonghaozheng64-commits:
Add two companion UI/backend extensions to the default custom-node catalog:

- **ComfyUI Prompt Workbench**: local prompt composition, field templates, saved character/outfit batches, pause/resume submission, Chinese/English controls and state backups.
- **ComfyUI LoRA Trigger Helper**: metadata backend for the workbench, offline safetensors scanning, editable groups and character output organization. Optional explicit online scans send SHA-256 hashes to Civitai. This is a backend-only extension with no standalone panel.

Both repositories are public and include installation/dependency documentation. They contain no model weights, generated assets or personal LoRA database. The helper starts with a local empty database. Its requirements.txt declares Send2Trash.

Validation: the upstream json-checker.py passes for custom-node-list.json; 11 workbench regression tests and 3 helper tests pass. The helper tests use temporary files and do not perform online scans or move existing outputs. Manager Local DB UI verification was not performed because a browser connection was unavailable.

Repositories:
- https://github.com/yonghaozheng64-commits/comfyui-prompt-workbench
- https://github.com/yonghaozheng64-commits/comfyui-lora-trigger-helper

PR #3273, opened 2026-09-12
fix(ui): separate enabled node conflicts from potential conflicts

### LINKED ISSUE #3092: Conflict detection uses global extension-node-map instead of locally installed extensions, causing inflated conflict count
## Description

The conflict detection feature in ComfyUI-Manager shows a conflict count that doesn't match the actual runtime conflicts. It compares against ALL known extensions in the global registry rather than just the locally installed ones.

## Steps to Reproduce

1. Install ComfyUI-Manager
2. Install a few custom nodes (e.g. comfyui-reactor-node, ComfyUI-WanVideoWrapper, etc.)
3. Open Manager -> Conflict Detection view
4. Observe the conflict count (e.g. showing 17)
5. Compare with the actual node list in ComfyUI's canvas

## Actual Result

The conflict detection view shows many more conflicts than actually exist at runtime. Conflicts are reported between extensions that the user doesn't even have installed.

## Expected Result

Conflict detection should only compare **currently installed extensions**, or at least clearly separate installed-vs-installed conflicts from installed-vs-global-catalog conflicts.

## Code Analysis

Looking at the source code in custom-nodes-manager.js (~lines 1260-1340):

The conflict detection flow:

1. loadNodes() calls the API endpoint \/v2/customnode/getmappings?mode=\\
2. This returns the data from \xtension-node-map.json\ cache file
3. This cache file contains ALL known extensions from the ComfyUI global registry, not just the locally installed ones
4. The code iterates through ALL entries and builds a \conflictsMap\:
   - Key: node class name
   - Value: list of extension keys that register this node
5. If a node class name appears in more than one extension entry, ALL those extensions get flagged as conflicting

The problem is that Step 2-3 uses the global extension catalog. If extension A and extension B in the global catalog both register a node called \CLIPTextEncode\ (a built-in ComfyUI node), they both get flagged as conflicting even if neither is installed locally.

## Suggested Fix

**Filter the extension-node-map to only include locally installed extensions before computing conflicts.**

In \custom-nodes-manager.js\, after fetching the mappings data, filter the entries to only include those whose URL matches a locally installed custom node folder (by checking folder names or git remotes in \custom_nodes\ directory).

Alternatively, add a separate API endpoint that returns mappings only for locally installed extensions.

## Environment

- ComfyUI-Manager version: 4.2.2
- ComfyUI: latest
- OS: Windows

---

*Reported after analyzing the source code of comfyui_manager js/custom-nodes-manager.js and the extension-node-map.json cache structure.*

yonghaozheng64-commits:
## Summary
Fixes #3092.

The legacy Manager UI builds node-name conflicts from the global catalog. This currently marks an enabled pack as conflicting even when its only peer is not installed or is disabled.

- Count conflicts only between packs whose backend `state` is `enabled`.
- Keep installation/enablement guidance: inactive packs display a separate **potential conflict** count and explicitly labeled peer list, considering enabled peers only.
- Preserve node listings and existing URL/title mapping resolution.
- Add eight dependency-free Node.js regression tests exercising the actual `loadNodes` method in an isolated VM with a mocked mappings response.

These remain catalog-based node-name overlaps, not proof of runtime registration collisions. No backend API, install permissions, dependencies, or local user data are changed.

## Validation
- `node --test tests/test_node_conflicts.cjs`: 8 passed.
- `python -m unittest discover -s tests -p "test_*.py"`: 45 passed.
- JavaScript module syntax check and `git diff --check`: passed.
- Running the tests against upstream main reproduces the enabled/uninstalled and enabled/disabled false positives (1 reported conflict instead of 0).
- Includes active/active, inactive/inactive, mixed catalog, duplicate node names, reload after state changes, missing catalog packs, and URL/title fallback cases.
- Full ComfyUI browser/end-to-end validation has not been performed.

Prepared with AI assistance and submitted for maintainer review. This branch is independent of plugin registration PR #3266.


yonghaozheng64-commits:
Hi maintainers, a gentle follow-up on this fix for #3092. Would you like any changes to the distinction between enabled-pack conflicts and potential conflicts, or any additional tests before review? The PR includes focused regression tests, but full ComfyUI browser/end-to-end validation has not been performed. Please let me know if there is a particular scenario you would like covered. Thank you for your time!

PR #3275, opened 2026-09-12
Added new custom nodes

domg73:
Added ComfyUI-LoadImageCrop and ComfyUI-LoadVideoCrop custom nodes


ltdrdata:
Thanks! Registered both entries (ComfyUI-LoadImageCrop and ComfyUI-LoadVideoCrop) directly because the list tail had drifted and the PR no longer merged cleanly — closing so the entries aren't added twice. They're now on the default channel.

PR #3277, opened 2026-09-13
Added my custom nodes to list

0xBeycan:


ltdrdata:
Thanks for the contribution! I went through the four repos in this PR — here's what I found.

**ComfyUI-WanAnimateLongVideoSampler** — clean, no issues found. These are pure diffusion compute nodes with no file writes, subprocess, exec, network calls, or custom routes.

**ComfyUI-WanAnimatePreprocess** — clean, no issues found. The pose/face preprocessor nodes load their models through the folder_paths helpers and add no custom routes or widget-driven file/exec sinks.

**ComfyUI-CaptionAudit** — one thing to fix before I can accept this one:

1. In nodes/audit.py, the `directory` widget flows through `resolve_dir` (line 103), which only applies abspath/expanduser with no base-directory containment, and `analyze` then reads `.txt` files under whatever path is given. Because the widget value is set by any workflow, this reads files from anywhere on the machine, and the file contents and names are then handed back through the `report_text`/`report_json` outputs (line 129), so the contents leak out too. Please confine the path: resolve it against a fixed base directory (realpath + commonpath against that base) and reject absolute paths and any `..` segments, so reads stay inside the intended folder.

**ComfyUI-BCNodes** — I couldn't grade this one because the staged tree had no analyzable source files. Could you double-check the repo contents so there's something to review?

0xBeycan:
Thanks for the review — that's helpful, and the CaptionAudit finding is fair.

On ComfyUI-BCNodes: sorry, that one was on me. I reset the repository while you were reviewing, so the tree you staged was empty. It's back up now with the full source.

More importantly, please don't merge this PR yet. I've ported all of my older node packs into BCNodes — CaptionAudit included — and I'm still working on it. The standalone CaptionAudit pack is now deprecated in favour of BCNodes, so the path-containment fix you asked for went into BCNodes rather than the old repo: the caption directory and the images directory are now resolved with realpath and confined to the ComfyUI tree (root, input/, output/, user/), with .. segments and symlinks collapsed before the check. Since caption datasets normally live outside the ComfyUI folder, extra roots can be allowed through a BC_CAPTION_ROOTS environment variable, which is read from the environment ComfyUI is started in and never from the workflow graph. A rejected path is shown on the node's error card instead of raising.

One thing I'd like your read on before you look at BCNodes properly: nodes/downloader.py. It registers five custom aiohttp routes under /bcnodes/downloader/* and downloads model files over http(s), and it stores Hugging Face / Civitai tokens on disk so gated files can be fetched. The destination is already constrained — every entry is resolved against folder_paths.models_dir with a commonpath check, only known model extensions are accepted and only http/https URLs pass — and the token endpoints never return the token values, just whether one is present. Is that shape acceptable for the registry, or do you want it restricted further (for example an allowlist of download hosts, or no server-side token storage at all)? I'd rather change it now than after a merge.

When I'm finished I'll update the JSON entry and push the final commit, then ping you for the re-review.

ltdrdata:
Thanks for the heads-up and for asking before merge — happy to hold until you ping me.

On nodes/downloader.py: keeping the tokens server side and never returning them from any route is the right approach, and the directory check against models/ is good. Two changes I'd ask for:

1. Confine the final file path (downloader.py:66-68, :100). When the file name comes from the URL, it is URL-decoded after the path is split, so an encoded separator ends up inside the name, and the joined path is never checked again — the file can land outside models/ (and .json / .yaml are among the allowed extensions, downloader.py:43). Please reject names containing "/", "\" or "..", and run realpath + commonpath against models/ on the final joined path, not only on the directory.

2. Restrict download hosts (downloader.py:80, :107-113). Any http(s) URL is accepted. A workflow someone shares can carry this node, and queuing it (downloader.py:438-463) or clicking Download in the first-open dialog — which shows the file name and target folder but not where it comes from (auto_model_downloader.js:460-463) — fetches from whatever server the workflow names, including machines on the user's local network. Please limit downloads to an explicit allow-list of hosts (Hugging Face, Civitai, and whatever else you really need) — the allow-list you suggested is exactly what I'd want. Please use one exact host check for both the allow-list and service_of, which decides where _auth sends the stored tokens (downloader.py:234-243): today it uses endswith("huggingface.co") / endswith("civitai.com"), so a lookalike domain also matches. Compare urlparse(url).hostname with host == "huggingface.co" or host.endswith(".huggingface.co"), and the same for civitai.com.

When BCNodes is ready and the JSON entry is final, ping me and I'll run the full review on the whole pack in one pass.

PR #3281, opened 2026-09-14
Add ComfyUI-Kavel to custom-node-list.json

hanshs474:
Adds [ComfyUI-Kavel](https://github.com/hanshs474/ComfyUI-Kavel): two nodes, **Kavel Generate Image** and **Kavel Edit Image**, that call the kavel.ai image API.

- Works out of the box without an API key (free tier), or with a key for account models
- No extra Python dependencies (stdlib + torch/numpy/Pillow already in ComfyUI)
- Edit node takes any `IMAGE` from the graph


ltdrdata:
Thanks for the contribution! I've registered `ComfyUI-Kavel` directly in `custom-node-list.json` — the list tail had drifted from your branch's base, so the PR couldn't be merged cleanly. Closing this so the entry isn't added twice; it's now available in ComfyUI-Manager.

PR #3284, opened 2026-09-15
Add Workflow Doctor to custom-node-list.json

lostmywind:
Adds [Workflow Doctor](https://github.com/lostmywind/comfyui-workflow-doctor) — a custom node that checks any workflow for the custom nodes **and model files** the local install is missing.

Why it complements Manager's own *Install Missing Custom Nodes*:

- it also covers **model files**, with HuggingFace / CivitAI search links for each one that is absent from `models/`
- it can check a workflow **before** you open it (paste JSON or pick a file from disk), not only the graph already on the canvas
- it installs nothing, downloads nothing and phones home to nobody — it is a read-only diff against the local install
- no dependencies beyond ComfyUI itself; also exposes a node and a sidebar panel

Validation: `python json-checker.py custom-node-list.json` → **Validation passed**. Diff is a single 11-line insertion at the end of the file.


ltdrdata:
Thanks for the contribution! I've registered `Workflow Doctor` directly in `custom-node-list.json` — the list tail had drifted from your branch's base, so the PR couldn't be merged cleanly. Closing this so the entry isn't added twice; it's now available in ComfyUI-Manager.

PR #3288, opened 2026-09-16
Add LumaFlux and ComfyUI-CachedSearch

shreshthsaini:
Registers two node packs.

**LumaFlux** (https://github.com/shreshthsaini/LumaFlux): SDR-to-HDR conversion with LumaFlux (arXiv 2604.02787). The paper repository doubles as the node pack (root `__init__.py` registers the nodes in `comfyui/`). A loader for the frozen FLUX.1-dev backbone (diffusers repo or the single-file checkpoints already in `models/diffusion_models` and `models/vae`) plus the released 71M-parameter adapters, a converter that runs the reference pipeline with video stabilization across a batch, an SDR preview, and save nodes for 16-bit PQ PNG (cICP-tagged) and 10-bit HEVC PQ video. `requirements.txt` holds the inference-only dependencies; the nodes import the package from `src/`.

**ComfyUI-CachedSearch** (https://github.com/shreshthsaini/ComfyUI-CachedSearch): best-of-N seed search (CachedSearch, arXiv 2607.23159). The drafts run under the native EasyCache node, an ImageReward verifier scores them, and only the winning seed is re-sampled at full compute. Sampler and verifier nodes; `requirements.txt` adds `image-reward`.

Both load cleanly with `Use local DB` checked. I maintain both repositories and am an author of both papers.


ltdrdata:
Thanks for the contribution! I've registered `ComfyUI-CachedSearch`, `LumaFlux` directly in `custom-node-list.json` — the list tail had drifted from your branch's base, so the PR couldn't be merged cleanly. Closing this so the entry isn't added twice; it's now available in ComfyUI-Manager.

PR #3292, opened 2026-09-16
Add "ComfyUI CoolB Nodes" to custom nodes list

CoolBreeze164:
Adds a pack of utility nodes for ComfyUI featuring saving/loading speakers with transcripts (useful alongside TTS packs), execution flow control (muting/bypassing nodes and groups on trigger to actively change the workflow state from run to run), some useful text file utilities (including save/load text file) and data type conversion.


ltdrdata:
Thanks for the contribution! I took a look at ComfyUI-CoolB-Nodes and ran into a couple of issues I'd like to see addressed before this can be merged.

1. In `TextFileWriter.save_file` (nodes.py:1520), the file path comes from a node widget, which any workflow can set, and it's joined to the ComfyUI root with `os.path.join` and no containment check — so an absolute or `../` path writes attacker-controlled content anywhere on the host.
2. In `TextFileReader.read_file` (nodes.py:1450), the file path is again taken from a node widget with no containment check, and the file's contents are returned as node output — so any path outside the intended folder can be read back.

Both stem from the same broken-containment issue at the path-construction layer. Please resolve each of these against a fixed base directory: use realpath and check with commonpath that the result stays inside that base, and reject absolute paths and any `..` segments.

CoolBreeze164:
Thank you for your review! I'm completely new to all this coding stuff, this was my first repo. You were right, after testing, the original writer and reader could indeed operate outside comfy root folder. I added a fix using Qwen:

- Absolute paths (`/etc/passwd`, `C:\Windows\...`, `\Windows\...`, `C:foo`, UNC) are rejected outright, so `os.path.join` can never drop the base.
- `..` segments are rejected up front, even ones that would resolve back inside the base.
- `realpath` + `commonpath` is the final proof: `realpath` collapses symlinks (so a symlink inside `output/` pointing outside is caught), and `commonpath` compares whole segments, avoiding the `startswith("/comfy/out")` vs `/comfy/outside` prefix bug. Cross-drive paths on Windows raise `ValueError` from `commonpath`, which is treated as an escape.
- `os.makedirs` now runs only after validation, so the writer can no longer create attacker-chosen directory trees outside the root.
- Both nodes keep their existing graceful behavior — violations surface as `Error: ...` node output instead of crashing the prompt.
- The `increment` branch only ever mutates the filename inside the already-validated directory, but it's re-checked with `_assert_within_base` anyway.

After testing the fix it seems to resolve the issues. If only comfy had text managers natively :(
Btw i'm planning to add another bigger pack soon. It would allow to make telegram bots inside comfy. Hopefully it won't have any issues :p

ltdrdata:
Thanks for the quick fix, and nice work — especially for a first repo! I re-checked the updated code, and the path handling itself is solid now: absolute paths and ".." are rejected, and the realpath + commonpath check keeps both nodes inside their base directory.

The remaining issue is which directory that base is. Both nodes use the whole ComfyUI root (nodes.py:1494 for the reader, nodes.py:1541 for the writer), and that is still too wide:

1. Save Text File (nodes.py:1539) can create or overwrite any file under the ComfyUI root, including Python files under custom_nodes/ that would run on the next start. Please set the base to a dedicated folder instead, e.g. the ComfyUI output directory (folder_paths.get_output_directory()) or a subfolder of it, and keep your existing containment check against that base.

2. Load Text File (nodes.py:1492) can read any file under the ComfyUI root and return its contents, which includes files in user/ that may hold settings or keys. Please confine it the same way, to the input or output directory (folder_paths.get_input_directory() / get_output_directory()).

Your _safe_resolve_under_base helper can stay exactly as it is — only the base_dir it is given needs to change.

CoolBreeze164:
Done! Changed base_dir to `ComfyUI/output`.
` _safe_resolve_under_base` still does the absolute/`..`/drive/UNC rejection plus the `realpath` + `commonpath` containment proof — it's just now proving containment against `<output>` instead of the root. Symlink escapes out of the output dir are still caught because both sides are realpathed.

ltdrdata:
Thanks, that's exactly it. I re-checked at `892e3c8`: both text nodes now resolve paths inside the output folder with your containment check, so this is registered.

One small thing: the category name and the display-name suffix contain a garbled emoji. `nodes.py:46` reads `CoolB вљЎпёЏ`, and the suffix in `__init__.py:29` is garbled the same way. It looks like the ⚡️ was saved through the wrong encoding, and it shows up that way in the ComfyUI menu. Re-saving those two lines as UTF-8 (or dropping the emoji) should fix it.

CoolBreeze164:
Thank you very much for your detailed guidance!❤️
Sorry for all this trouble with such a simple node pack. I will avoid these mistakes i made in the future.

PR #3295, opened 2026-09-18
Add ComfyUI Custom Node Manager to custom-node-list

Raykosan:
### New node

**Name:** ComfyUI Custom Node Manager
**Repo:** https://github.com/Raykosan/ComfyUI-Custom-Node-Manager

### What it does

An alternative manager for `custom_nodes` that bypasses ComfyUI Manager
restrictions. Works entirely via git and the GitHub REST API — it does
**not** contact `api.comfy.org` at all.

### Features

- Install any node from a Git URL (with optional tag/branch/commit pin)
- Switch Version: browse and switch to any tag, **including versions flagged by Comfy Registry**
- Update checking via `git ls-remote` (no API rate limits)
- Batch update with checkboxes
- Auto-stash + stash management UI
- Attach Git Remote for zip-installed nodes (no `.git`)
- Extended version detection (metadata.json / pyproject.toml / `__version__` / setup.py / package.json / git tags)
- i18n: EN / 中文 / Русский
- Optional `GITHUB_TOKEN` for higher rate limits

### Why

Comfy Registry's YARA-based security scanner produces frequent false
positives. Flagged versions are hidden from "Select Version" in the
official manager, leaving users stuck on outdated versions even when
authors release critical fixes. This extension gives users control back
by working directly with git, not with the Registry.

### Checklist

- [x] `__init__.py` exports `WEB_DIRECTORY` and `NODE_CLASS_MAPPINGS`
- [x] `requirements.txt` present (`requests`, `tomli` for Python < 3.11)
- [x] `pyproject.toml` present
- [x] `README.md` present
- [x] `LICENSE` present (Apache 2.0)
- [x] Tested locally with Python 3.13 / Windows
- [x] No changes outside `node_db/new/custom-node-list.json`


ltdrdata:
Thanks for the contribution!

I took a look and I can't merge this as it stands. The extension exposes a set of open web routes with no access control, so any remote caller can reach them, and each one performs a powerful action on the host from a value the caller supplies:

1. POST /custom_node_manager/install (CustomNode.py:585) takes any caller-supplied git_url, clones it, and runs pip install on its requirements.txt — this runs code the caller supplies on the machine. Please make this install action local-only / explicitly user-initiated rather than reachable over the network, or remove the route; if it must stay, restrict git_url to an explicit allow-list of hosts.

2. POST /custom_node_manager/update (CustomNode.py:628) takes a caller-supplied folder and version, checks the version out, and re-runs pip install with no origin check — same concern. Please make this local-only / user-initiated, or remove the route.

3. POST /custom_node_manager/restart (CustomNode.py:717) calls os.execv to restart the process on request. Please make this local-only / explicitly user-initiated, not reachable over the network.

4. POST /custom_node_manager/remove (CustomNode.py:656) deletes a caller-named node directory via shutil.rmtree. Please resolve the target against a fixed base directory (realpath + commonpath against that base, rejecting absolute paths and ".." segments) and make the removal explicitly user-initiated rather than openly reachable.

5. POST /custom_node_manager/attach_remote (CustomNode.py:878) runs git init/remote/fetch/reset against a caller-supplied git_url on an existing folder with no origin check. Please restrict git_url to an explicit allow-list of hosts, or make this local-only.

These routes share the same shape — an open route acting on a value the caller supplies — so the cleanest fix is to make the install / update / restart / remove / attach_remote actions local-only and explicitly user-initiated rather than reachable over the network, and to confine any path or host they take (a fixed base directory for the folder in remove, an allow-list for the git_url in install and attach_remote).

Raykosan:
> Thanks for the contribution!
> 
> I took a look and I can't merge this as it stands. The extension exposes a set of open web routes with no access control, so any remote caller can reach them, and each one performs a powerful action on the host from a value the caller supplies:
> 
>     1. POST /custom_node_manager/install (CustomNode.py:585) takes any caller-supplied git_url, clones it, and runs pip install on its requirements.txt — this runs code the caller supplies on the machine. Please make this install action local-only / explicitly user-initiated rather than reachable over the network, or remove the route; if it must stay, restrict git_url to an explicit allow-list of hosts.
> 
>     2. POST /custom_node_manager/update (CustomNode.py:628) takes a caller-supplied folder and version, checks the version out, and re-runs pip install with no origin check — same concern. Please make this local-only / user-initiated, or remove the route.
> 
>     3. POST /custom_node_manager/restart (CustomNode.py:717) calls os.execv to restart the process on request. Please make this local-only / explicitly user-initiated, not reachable over the network.
> 
>     4. POST /custom_node_manager/remove (CustomNode.py:656) deletes a caller-named node directory via shutil.rmtree. Please resolve the target against a fixed base directory (realpath + commonpath against that base, rejecting absolute paths and ".." segments) and make the removal explicitly user-initiated rather than openly reachable.
> 
>     5. POST /custom_node_manager/attach_remote (CustomNode.py:878) runs git init/remote/fetch/reset against a caller-supplied git_url on an existing folder with no origin check. Please restrict git_url to an explicit allow-list of hosts, or make this local-only.
> 
> 
> These routes share the same shape — an open route acting on a value the caller supplies — so the cleanest fix is to make the install / update / restart / remove / attach_remote actions local-only and explicitly user-initiated rather than reachable over the network, and to confine any path or host they take (a fixed base directory for the folder in remove, an allow-list for the git_url in install and attach_remote).
> 
> I'll re-evaluate once these are addressed.

Thanks for the thorough review — all points are valid.

I've addressed them in the main repository at v0.3.0:
https://github.com/Raykosan/ComfyUI-Custom-Node-Manager/releases/tag/0.3.0

Summary of fixes:

1. **Local-only enforcement.** Added `security.py` with a `local_only`
   decorator applied to every route. Requests from non-loopback
   addresses are rejected with 403. Uses `request.remote` +
   `ipaddress.is_loopback`, not Origin/Referer.

2. **Session token for POST routes.** A `secrets.token_urlsafe(32)`
   is generated at startup. Every state-changing POST route requires
   the `X-CNM-Token` header, compared with `secrets.compare_digest`.
   The token is served only from `GET /custom_node_manager/token`,
   which is itself local-only — so remote clients cannot obtain it.

3. **git_url allow-list.** `install` and `attach_remote` validate
   URLs against `{github.com, gitlab.com, bitbucket.org, codeberg.org,
   gitea.com}`. Overridable via `CNM_ALLOWED_GIT_HOSTS` env var.
   Both HTTP(S) and SSH forms (`git@host:owner/repo`) are parsed.

4. **remove path confinement.** Rejects absolute paths, path
   separators, and `..` segments. Then uses
   `os.path.commonpath([realpath(root), realpath(target)]) == realpath(root)`
   against every `custom_nodes` root, and refuses to remove a root itself.

5. **`restart`** now requires both local-only + token, like every
   other POST route.

The extension remains fully functional via the UI: the browser
obtains the token once and attaches it to every action. Remote
callers cannot reach any of the mutating routes.

Reverified and re-tagged. Thanks again.

Raykosan:
Good catch — this is a real DNS rebinding vector. Confirmed and fixed in v0.3.1.

**The issue:**

`GET /token` checked only `request.remote` (TCP peer = 127.0.0.1). A rebinding page could resolve its own domain to 127.0.0.1, read the token, then POST /install with an attacker repo — RCE.

**The fix:**

`is_local_request` now also validates the `Host` header. A request must have **both** a loopback TCP peer **and** a loopback `Host` value. Since browsers don't allow JS to spoof `Host`, a rebinding page will still carry `Host: evil.com` and be rejected with 403.

```python
def is_local_request(request) -> bool:
    # 1. TCP peer must be loopback
    remote = request.remote
    if not _is_loopback_ip(remote):
        return False
    # 2. Host header must also be loopback (DNS rebinding protection)
    if not _is_loopback_host(request.host or ""):
        return False
    return True
```
`_is_loopback_host` accepts localhost, IPv4 127.0.0.0/8, and IPv6 [::1] (with or without port).

Verified locally:

- curl http://127.0.0.1:8188/custom_node_manager/token → 200 + token

- curl http://localhost:8188/custom_node_manager/token → 200 + token

- curl --resolve evil.com:8188:127.0.0.1 http://evil.com:8188/custom_node_manager/token → 403 {"error":"local requests only"}

Tagged as v0.3.1: https://github.com/Raykosan/ComfyUI-Custom-Node-Manager/releases/tag/0.3.1

PR #3304, opened 2026-09-20
Match KayTool's URL casing to its registry entry

kk8bit:
`custom-node-list.json` lists KayTool as `https://github.com/kk8bit/KayTool`, but the Comfy Registry reports the repository as `https://github.com/kk8bit/kaytool` (the canonical GitHub name is lowercase; the capitalised form only works because GitHub redirects).

`git_utils.normalize_url()` does not fold case, so the two strings never match:

```python
>>> git_utils.normalize_url('https://github.com/kk8bit/KayTool')
'https://github.com/kk8bit/KayTool'
>>> git_utils.normalize_url('https://github.com/kk8bit/kaytool')
'https://github.com/kk8bit/kaytool'
```

`repo_cnr_map` is keyed by that normalised URL, so the registry entry and the list entry are never paired and KayTool shows up as two separate nodes in the manager. Neither one installs:

- the cnr entry has no `files` key, so `install_by_id` raises `KeyError: 'files'` at `manager_core.py:1499`
- the git-clone entry isn't recognised as a registry node, so it's refused with `This installation is not allowed in this security_level`

Users have been hitting this for months (kk8bit/kaytool#18, kk8bit/kaytool#23, kk8bit/kaytool#34, kk8bit/kaytool#35). I'm the author of KayTool.

This PR just corrects the casing in `custom-node-list.json` so the entry matches the registry.

As a separate thought: GitHub owner and repository names are case-insensitive, so any node whose list URL differs in case from its registry `repository` field will hit the same thing. Lowercasing the GitHub id inside `normalize_url()` would make this class of mismatch impossible — happy to open a second PR for that if you'd like it.


kk8bit:
One more consequence of the casing mismatch, for context on why this matters beyond the duplicate entry.

With the current list, `getlist` returns KayTool as two unrelated packs:

| entry | source | version | stars | trust |
|---|---|---|---|---|
| `KayTool` | custom-node-list (git-clone, `…/KayTool`) | `unknown` | 380 | true |
| `kaytool` | registry (cnr, `…/kaytool`) | 0.71.1 | **-1** | **false** |

- The list entry sorts to the bottom because `custom-nodes-manager.js` puts every `version == 'unknown'` row last.
- The registry entry's `reference` is `https://registry.comfy.org/nodes/kaytool`, which `populate_github_stats()` can't find in `github-stats.json` (keyed by the capitalised GitHub URL), so it gets `stars = -1` / `trust = false` and sinks too.

So a 380-star pack with an Active registry release is currently shown twice, once without a version and once as untrusted, and both copies rank near the bottom. Lower-casing the list URL lets `repo_cnr_map` merge them into one entry that carries the version from the registry and the stats from the list.

Happy to open a follow-up for case-insensitive matching in `normalize_url` if you'd prefer to fix the class of problem rather than this instance.

kk8bit:
Closing this one: I've aligned it from my side instead. KayTool 0.71.2 publishes with `Repository = https://github.com/kk8bit/KayTool` in `pyproject.toml`, so the registry's `repository` field now matches the list entry as written and `repo_cnr_map` merges the two without any change here.

The offer above stands — if case-insensitive matching in `normalize_url` would be welcome, I'll send it as its own PR, since any other pack whose list URL differs in case from its registry metadata hits the same split.

